PR #80, opened 2022-02-18
Add MOCMC

brryan:
First PR for:
1) Adding MOCMC sample transport
2) Interfacing MOCMC outputs (Eddington tensor and frequency-averaged opacities) with existing radiation moments

Some other notes:
- [X] Converted moment fluid sources to Rosseland mean opacities from new `singularity-opac` feature
- [X] Replaced `B_fake` with a scale-free thermal distribution option for gray opacities in `singularity-opac` -- basically it's just `c = kb = h = 1` so `Bnu` still has structure in frequency. 
- [X] Switch to scale-free radiation distribution + gray opacity in singularity, `Bnu = 2 Ns nu^3 / (Exp(nu / T) + 1)`
- [X] Add MOCMC versions of advection and equilibration tests
- [X] Merge https://github.com/lanl/parthenon/pull/653 into `develop` and update parthenon submodule here
- [ ] Merge https://github.com/lanl/parthenon/pull/679 into `develop` and update parthenon submodule here


src/radiation/moments.cpp
```diff
@@ -91,29 +93,34 @@ TaskStatus MomentCon2PrimImpl(T* rc) {
...
         c.GetCovTilPiFromCon(E, covF, xi, phi, &conTilPi);
-        if (STORE_GUESS) { 
+        if (STORE_GUESS) {
           v(b, iXi(ispec), k, j, i) = xi;
           v(b, iPhi(ispec), k, j, i) = phi;
         }
-        c.Con2Prim(E, covF, conTilPi, &J, &covH); 
+        if (EDDINGTON_KNOWN) {
+          SPACELOOP2(ii, jj) { conTilPi(ii, jj) = v(b, iTilPi(ispec, ii, jj), k, j, i); }
+          c.Con2Prim(E, covF, conTilPi, &J, &covH);
```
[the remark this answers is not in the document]
brryan: Works great, thanks :+1:

src/fluid/fluid.cpp
```diff
@@ -567,10 +567,10 @@ TaskStatus CalculateFluidSourceTerms(MeshBlockData<Real> *rc,
...
           src(ceng, k, j, i) = gdet * alpha * (Ta - TGam);
 #else
-          SPACETIMELOOP2(mu, nu) {
-            TGam += Tmunu[mu][nu] * gam[nu][0][mu];
-          }
-          src(ceng,k,j,i) = gdet * TGam;
+                         SPACETIMELOOP2(mu, nu) {
+                           TGam += Tmunu[mu][nu] * gam[nu][0][mu];
+                         }
+                         src(ceng, k, j, i) = gdet * TGam;
```
brryan, line 573: I don't know what's going on with `clang-format-12` here

inputs/mocmc_advection.pin
```diff
@@ -0,0 +1,120 @@
...
+gray_kappa = 1.e3
+
+<radiation_advection>
+J = 1.0
+Hx = 0.0
+Hy = 0.0
+Hz = 0.0
+vx = 0.3
+width = 0.0333
+kappas_init = 1.e3
```
brryan, line 119:
```suggestion
tau = 1.e3
```

scripts/plot_thermalization.py
```diff
@@ -0,0 +1,178 @@
...
+    Inu[n] += dInu[n]
+    Inu[n] = max(Inu[n], 1.e-100)
+
+  for n in range(nnu - 1):
+    dE += 4.*np.pi/cgs['CL']*(dInu[n+1] + dInu[n])/2*(nus[n+1]-nus[n])
+
+  E[cycle+1] = E[cycle] + dE
+  u[cycle+1] = u[cycle] - dE
+  Tr[cycle+1] = pow(15.*cgs['CL']**3*cgs['HPL']**3*E[cycle+1]/(7.*cgs['KBOL']**4*np.pi**5*NSPECIES),1./4.)
+  Tg[cycle+1] = u[cycle+1]*mu*(gamma - 1.)/(rho0*cgs['KBOL'])
```
brryan, line 153: @lroberts36 This script calculates zero-D gas/radiation thermalization with frequency dependence by evolving the radiation spectrum. Could be usefully combined with your thermalization pgen test, I think

src/main.cpp
```diff
@@ -46,32 +47,7 @@ int main(int argc, char *argv[]) {
   // pman.app_input->UserWorkAfterLoop = phoebus::UserWorkAfterLoop;
   // pman.app_input->SetFillDerivedFunctions = phoebus::SetFillDerivedFunctions;
 
-  // TODO(JMM): Move this into another function somewhere?
```
brryan, line 49: Moved :)

src/microphysics/eos_phoebus/eos_phoebus.cpp
```diff
@@ -125,14 +125,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
 
-  params.Add("unit_conv", phoebus::UnitConversions(pin));
-  auto &unit_conv = params.Get<phoebus::UnitConversions>("unit_conv");
-
-  // Store unit conversions mainly for dump file output
-  params.Add("length_unit", unit_conv.GetLengthCodeToCGS());
-  params.Add("time_unit", unit_conv.GetTimeCodeToCGS());
-  params.Add("mass_unit", unit_conv.GetMassCodeToCGS());
-  params.Add("temperature_unit", unit_conv.GetTemperatureCodeToCGS());
+  phoebus::UnitConversions unit_conv(pin);
```
brryan, line 128: Units moved to `phoebus` package

src/pgen/radiation_advection.cpp
```diff
@@ -103,7 +146,44 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
+  if (radpkg->Param<bool>("active")) {
+    if (radpkg->Param<std::string>("method") == "mocmc") {
+      radiation::MOCMCInitSamples(rc.get());
+    }
+  }
+
+  radiation::MomentPrim2Con(rc.get(), IndexDomain::interior);
+}
+
+void ProblemModifier(ParameterInput *pin) {
```
brryan, line 160: At the moment I'm modifying some constants so `moment_*` always runs in scale-free mode and `mocmc` (and `monte_carlo`, though untested) always runs in scale mode. These can maybe be unified with an appropriately scale-free radiation distribution function inside `singularity-opac` in the (near?) future.

src/phoebus_boundaries/phoebus_boundaries.cpp
```diff
@@ -189,6 +189,22 @@ void ReflectOuterX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
...
+  return DeviceAllocate<ParticleBoundIX1Outflow>();
+}
+std::unique_ptr<parthenon::ParticleBound,
+                parthenon::DeviceDeleter<parthenon::DevMemSpace>>
+SetSwarmOX1Outflow() {
+  return DeviceAllocate<ParticleBoundOX1Outflow>();
+}
+std::unique_ptr<parthenon::ParticleBound,
+                       parthenon::DeviceDeleter<parthenon::DevMemSpace>>
+SetSwarmNoWorkBC() {
```
brryan, line 204: For MOCMC particles, the boundaries generically do things like look at the fluid state to compute thermal spectra which isn't possible with the existing parthenon boundaries. So I sidestep those boundary calls with this `NoWork` function and update the MOCMC particles at boundaries in a separate kernel in `phoebus`.

src/phoebus_utils/programming_utils.hpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+template <typename T>
+bool ContainedInVector(T &item, std::vector<T> &container) {
+  if (std::find(container.begin(), container.end(), item) == container.end()) {
+    std::stringstream msg;
+    return false;
+  }
+  return true;
+}
```
brryan, line 45: Others may have other generally useful little gizmos worth putting in here? One nice thing might be a (host-side) class that maps between strings and enums, for e.g. safely converting from string inputs to internal enums.

src/phoebus_utils/root_find.hpp
```diff
@@ -18,6 +18,44 @@ namespace root_find {
 
 enum class RootFindStatus { success, failure };
 
+template <typename F>
+KOKKOS_INLINE_FUNCTION void broyden2(F &func, Real v[2], const int max_iter = 50,
```
brryan, line 22: Not fully implemented/tested but we would want something like this if we do 2D rootfinds over T/Ye for the radiation a la Fornax.

src/phoebus_utils/unit_conversions.hpp
```diff
@@ -58,6 +58,38 @@ class UnitConversions {
...
+    const Real LENGTH = unit_conv.GetLengthCGSToCode();
+    const Real ENERGY = MASS * LENGTH * LENGTH / (TIME * TIME);
+    const Real TEMPERATURE = unit_conv.GetTemperatureCGSToCode();
+    h_code_ = pc::h * MASS * LENGTH * LENGTH / TIME;
+    c_code_ = pc::c * LENGTH / TIME;
+    kb_code_ = pc::kb * ENERGY / TEMPERATURE;
+    mp_code_ = pc::mp * MASS;
+  }
+
+  Real h() const { return h_code_; }
```
brryan, line 78: constants in code units show up occasionally in phoebus radiation right now, e.g. going from frequency in code units to neutrino energy in code units.

src/phoebus_utils/programming_utils.hpp
```diff
@@ -0,0 +1,49 @@
...
+struct is_specialization_of<TEMPL<TPARAMS...>, TEMPL> : public std::true_type {};
+
+template <typename T>
+bool soft_equiv(const T x, const T y, const T tolerance = static_cast<T>(1.e-10f)) {
+  bool equivalent = false;
+  if (2. * fabs(y - x) < tolerance * (std::fabs(x) + std::fabs(y))) {
+    equivalent = true;
+  }
+
+  return equivalent;
```
Yurlungur, line 35: Why not just
```suggestion
  return (2. * fabs(y - x) < tolerance * (std::fabs(x) + std::fabs(y)));
```

Yurlungur: also definitely nice to have this.

src/phoebus_utils/programming_utils.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <algorithm>
+#include <vector>
+
+namespace programming {
+
+template <class SPECIAL, template <class...> class TEMPL>
+struct is_specialization_of : public std::false_type {};
+
+template <template <class...> class TEMPL, class... TPARAMS>
+struct is_specialization_of<TEMPL<TPARAMS...>, TEMPL> : public std::true_type {};
```
Yurlungur, line 26: I understand what this is for and why it has utility. But... YUCK.

src/phoebus_utils/programming_utils.hpp
```diff
@@ -0,0 +1,49 @@
...
+    equivalent = true;
+  }
+
+  return equivalent;
+}
+
+template <typename T>
+bool ContainedInVector(T &item, std::vector<T> &container) {
+  if (std::find(container.begin(), container.end(), item) == container.end()) {
+    std::stringstream msg;
```
Yurlungur, line 41: Why is this stringstream here?

src/phoebus_utils/programming_utils.hpp
```diff
@@ -0,0 +1,49 @@
...
+  bool equivalent = false;
+  if (2. * fabs(y - x) < tolerance * (std::fabs(x) + std::fabs(y))) {
+    equivalent = true;
+  }
+
+  return equivalent;
+}
+
+template <typename T>
+bool ContainedInVector(T &item, std::vector<T> &container) {
```
Yurlungur, line 39: What is the use case for this? If you want to make calls like this regularly, maybe a `std::vector` is the wrong container and a set or map is better?

src/phoebus_utils/programming_utils.hpp
Yurlungur: :+1: for programming gizmos. Wouldn't a class that did that have to be enum specific? Maybe not if we played evil macro games...

src/phoebus_utils/root_find.hpp
Yurlungur: :+1: If I recall, @lroberts36 has something like this in the moment code too right? Or is that a different-dimensional root find?

src/phoebus_utils/root_find.hpp
```diff
@@ -18,6 +18,44 @@ namespace root_find {
...
+    Real newresid[2] = {0.};
+    func(v[0], v[1], newresid);
+    Real z[2] = {newresid[0] - resid[0], newresid[1] - resid[1]};
+
+    // J = J + (outer((z - J.s),s)) / ( s.s )
+    for (int ii = 0; ii < 2; ii++) {
+      resid[ii] = newresid[ii];
+    }
+
+    niter++;
```
Yurlungur, line 47: Do we want a line search in here?

src/phoebus_utils/unit_conversions.hpp
Yurlungur: All this should be `KOKKOS_FORCEINLINE_FUNCTION`. Except the constructor.

Although actually, is there any reason this class can't be a struct? Just make all the code units vars public and get rid of the getter functions. It's a simple enough little thing that that makes sense to me.

src/radiation/closure.hpp
```diff
@@ -79,7 +79,6 @@ class ClosureEdd {
 
   //-------------------------------------------------------------------------------------
   /// Calculate the momentum density flux P^i_j from J, \tilde H_i, and \tilde pi^{ij}.
-  KOKKOS_FUNCTION
```
Yurlungur, line 82: Why does this not need to be a `KOKKOS_FUNCTION`?

src/radiation/closure_mocmc.hpp
```diff
@@ -0,0 +1,70 @@
...
+  KOKKOS_FUNCTION
+  ClosureMOCMC(const Vec con_v_in, LocalGeometryType *g)
+      : ClosureEdd<Vec, Tens2, SET>(con_v_in, g) {}
+
+  KOKKOS_FUNCTION
+  ClosureStatus GetCovTilPiFromPrim(const Real J, const Vec cov_H, Tens2 *con_tilPi) {
+    PARTHENON_FAIL("MOCMC has already stored TilPi elsewhere!");
+
+    // Do nothing
+    return ClosureStatus::success;
```
Yurlungur, line 56: `ClosureStatus::fail` since you shouldn't be doing this call?

Perhaps this implies we should rethink the closure base class?

src/radiation/kd_grid.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include "geometry/geometry.hpp"
+
+namespace radiation {
+namespace kdgrid {
+
+KOKKOS_INLINE_FUNCTION
+void integrate_ninj_domega_quad(const Real xi0, const Real xi1, const Real phi0,
+                                const Real phi1, Real *wgts) {
+  // expressions validated through numerical quadrature in python
```
Yurlungur, line 32: can we put that python script somewhere in the codebase?

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,1100 @@
...
+  auto pJ = imap.GetFlatIdx(pr::J);
+  auto pH = imap.GetFlatIdx(pr::H);
+  auto pdens = imap[pf::density].first;
+  auto pv = imap.GetFlatIdx(fluid_prim::velocity);
+  auto pT = imap[pf::temperature].first;
+  auto pye = imap[pf::ye].first;
+  auto dn = imap[im::dnsamp].first;
+
+  const int nblock = v.GetDim(5);
+  PARTHENON_REQUIRE_THROWS(nblock == 1, "Packing not currently supported for swarms");
```
Yurlungur, line 101: You could just make this function only take `MeshBlockData` rather than template it for now.

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,1100 @@
...
+namespace ir = radmoment_internal;
+namespace im = mocmc_internal;
+using namespace singularity;
+using namespace singularity::neutrinos;
+using singularity::RadiationType;
+using vpack_types::FlatIdx;
+
+constexpr int MAX_SPECIES = 3;
+
+// TODO(BRR) add options
```
Yurlungur, line 46: What's the plan for the optionality? Not that we need to think too hard about that for now.

Yurlungur: One reason I ask is I'm wondering how we will pass those settings through into the kernel.

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,1100 @@
...
+KOKKOS_INLINE_FUNCTION int get_nsamp_per_zone(const int &k, const int &j, const int &i,
```
Yurlungur, line 47: `FORCEINLINE`

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,1100 @@
...
+  int index = 0;
+  for (int k = 0; k < nx_k; k++) {
+    for (int j = 0; j < nx_j; j++) {
+      for (int i = 0; i < nx_i; i++) {
+        starting_index(k, j, i) = index;
+        index += static_cast<int>(dN_h(k + kb.s, j + jb.s, i + ib.s));
+      }
+    }
+  }
+  starting_index.DeepCopy(starting_index_h);
```
Yurlungur, line 141: This looks like it could be very slow for large meshblocks. I wonder if there's a way to do this kind thing in parallel or on device? Probably not. But maybe worth engaging with the Kokkos people to see if they have a solution.

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,1100 @@
...
+        Real cov_g[4][4];
+        geom.SpacetimeMetric(CellLocation::Cent, k, j, i, cov_g);
+        Real alpha = geom.Lapse(CellLocation::Cent, k, j, i);
+        Real con_beta[3];
+        geom.ContravariantShift(CellLocation::Cent, k, j, i, con_beta);
+        const Real vpcon[] = {v(pv(0), k, j, i), v(pv(1), k, j, i), v(pv(2), k, j, i)};
+        const Real W = phoebus::GetLorentzFactor(vpcon, cov_g);
+        const Real ucon[4] = {W / alpha, vpcon[0] - con_beta[0] * W / alpha,
+                              vpcon[1] - con_beta[1] * W / alpha,
+                              vpcon[2] - con_beta[2] * W / alpha};
```
Yurlungur, line 176: wonder if this should be a relativity utils function

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,1100 @@
...
+        for (int n = 0; n < nsamp; n++) {
+          const int nswarm = swarm_d.GetFullIndex(k, j, i, n);
+
+          Real nu_fluid0 = nusamp(0);
+          Real nu_lab0 = 0.;
+          SPACETIMELOOP(nu) { nu_lab0 -= ncov(nu, nswarm) * ucon[nu]; }
+          nu_lab0 *= nusamp(0);
+
+          const Real shift = (log(nusamp(0)) - log(nu_lab0)) / dlnu;
+          interpolation::PiecewiseConstant interp(nu_bins, dlnu, shift);
```
Yurlungur, line 933: Really feels like we want at least piecewise linear here.

Yurlungur (approved):
Really really exciting to get this in, @brryan ! I have to admit, I only skimmed the dense technical section, so @jdolence and @lroberts36 should also look. None of my comments should be considered blocking.

src/fluid/fluid.cpp
brryan: Ugh I really hate doing that... maybe as code changes this formatting bug will go away later

src/phoebus_boundaries/phoebus_boundaries.cpp
brryan: Right now the parthenon particles boundaries are written differently from the cell-based boundaries -- the particle boundary calls are inherited classes holding `KOKKOS_INLINE_FUNCTION` functions that are called inline by the `Swarm` internally. The function signature is fixed so I don't think there's an easy way to get application-specific variables out. This is to avoid always calling like six separate kernels over all particles. Maybe there is a better way to do this but I think the issue is more specific to my `Swarm` coding than to the more general parthenon boundaries stuff.

src/phoebus_utils/programming_utils.hpp
brryan: I was thinking some kind of class that held like an `std:map<MyEnum, std::string>` internally or something. Maybe that doesn't work, I didn't work it all the way out.

src/phoebus_utils/root_find.hpp
brryan: I think that's an analytic inversion but maybe I'm mistaken.

src/phoebus_driver.cpp
```diff
@@ -87,6 +87,57 @@ TaskListStatus PhoebusDriver::Step() {
   return status;
 }
 
+// TODO(BRR) This is required for periodic BCs, unless the issue is radiation not being
+// included in ConvertBoundaryConditions
```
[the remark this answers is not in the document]
brryan: This is only called on initialization -- if I don't call this, periodic BCs dont seem to update ghost zones between problem generation and the first timestep. It's something that came up when I switched from calling the ProblemGenerator over `interior` rather than `entire` zones. Maybe a parthenon issue? I haven't completely closed the loop on this. But again this is only before timestepping begins, comms arent serialized for radiation during time cycling.

src/phoebus_driver.cpp
```diff
@@ -244,6 +317,47 @@ TaskCollection PhoebusDriver::RungeKuttaStage(const int stage) {
...
+                                         std::cout << "MOCMC total particles = "
+                                                   << static_cast<long>((*res)[0])
+                                                   << std::endl;
+                                         return TaskStatus::complete;
+                                       },
+                                       &particle_resolution.val)
+                                 : none);
+    }
+  }
+
```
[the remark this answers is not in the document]
brryan: Yeah in principle I would like to break this up I think but a lot of this can also be woven together pretty tightly... I'm not sure what the optimal solution is. We may not be sure what it is until all the code is already written...

src/phoebus_utils/programming_utils.hpp
brryan: It actually works really well. But I have no idea what this code is doing.

src/radiation/mocmc.cpp
brryan: I'm not too sure what options we want beyond this are. But eventually I think we'll want multiple functions, including ones more sophisticated than just the uniform-over-all-zones function here.

src/radiation/mocmc.cpp
brryan: Yeah I would like to find a better way to do this (it's an issue inside the parthenon particles as well)

src/radiation/mocmc.cpp
brryan: Yeah I should move it there :+1:

src/phoebus_utils/programming_utils.hpp
lroberts36: It is taken almost directly from this standard lib proposal: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2098r0.pdf

src/radiation/mocmc.cpp
jdolence: isn't this just a prefix sum which you can do with a `Kokkos::parallel_scan`?

src/radiation/mocmc.cpp
brryan: Yeah piecewise constant is yucky. I remember having some issues previously with scattering where the interpolations to the fluid frame and then back to the coordinate frame weren't exactly cancelling, and so that disagreement became a larger error in the sample spectra relative to the actual problem evolution. Something to look into in the (near) future.

src/phoebus_utils/unit_conversions.hpp
brryan: Doing that while keeping the vars `const` would require putting the whole constructor in the initializer list which in my opinion would be unreadable. Maybe theres another way?

scripts/constants.py
```diff
@@ -0,0 +1,33 @@
+from math import pi
```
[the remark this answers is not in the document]
brryan: Yes, done

scripts/plot_thermalization.py
```diff
@@ -0,0 +1,178 @@
...
+import numpy as np
+import matplotlib
+import matplotlib.pyplot as plt
+import matplotlib.patches as patches
+import sys
+import shutil
+import os
+from subprocess import call, DEVNULL
+import glob
+from parthenon_tools import phdf
```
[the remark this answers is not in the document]
brryan: Yeah this only generates the analytic solution for comparison right now, but eventually it will read in the phoebus solution for comparison

scripts/python/plot_radiation_advection.py
```diff
@@ -14,36 +14,36 @@
 # publicly, and to permit others to do so.
```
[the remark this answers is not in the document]
brryan: Already there

src/phoebus_boundaries/phoebus_boundaries.cpp
```diff
@@ -189,6 +189,22 @@ void ReflectOuterX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
       });
 }
 
+std::unique_ptr<parthenon::ParticleBound,
+                parthenon::DeviceDeleter<parthenon::DevMemSpace>>
```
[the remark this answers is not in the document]
brryan: `ambiguating new declaration` issue when compared with what's in the header file (which can't be `auto` I don't think). But these tiny functions can just be marked `inline` and put in the header file, which I just did

src/phoebus_boundaries/phoebus_boundaries.cpp
```diff
@@ -231,4 +247,75 @@ TaskStatus ConvertBoundaryConditions(std::shared_ptr<MeshBlockData<Real>> &rc) {
...
+          Boundaries::OutflowInnerX1;
+      if (rad_method == "mocmc") {
+        pman.app_input->swarm_boundary_conditions[parthenon::BoundaryFace::inner_x1] =
+            Boundaries::SetSwarmNoWorkBC;
+      } else {
+        pman.app_input->swarm_boundary_conditions[parthenon::BoundaryFace::inner_x1] =
+            Boundaries::SetSwarmIX1Outflow;
+      }
+    } // else, parthenon periodic boundaries
+  }
```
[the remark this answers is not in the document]
brryan: Yeah this is yucky but I'm not immediately sure how to improve it

src/phoebus_driver.hpp
```diff
@@ -43,6 +45,9 @@ class PhoebusDriver : public EvolutionDriver {
   AllReduce<int> particles_outstanding;
 };
 
+// Simple package for storing simulation- and problem-wide objects and parameters for I/O
+std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin);
+
```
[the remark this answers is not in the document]
brryan: Good idea, done

src/phoebus_utils/interpolation.hpp
```diff
@@ -0,0 +1,89 @@
...
+// rights in the program are reserved by Triad National Security, LLC,
+// and the U.S. Department of Energy/National Nuclear Security
+// Administration. The Government is granted for itself and others
+// acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works,
+// distribute copies to the public, perform publicly and display
+// publicly, and to permit others to do so.
+
+#ifndef PHOEBUS_UTILS_INTERPOLATION_
+#define PHOEBUS_UTILS_INTERPOLATION_
```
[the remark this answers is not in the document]
brryan: From our earlier discussion it sounds like our interpolation use cases are sufficiently different (grid vs arbitrary 1D uniform data) that unification wouldn't be trivial. My suggestion is that we implement what we need and when everything's in look at it and see whether things should be combined. We do now share the namespace `interpolation` I believe.

src/phoebus_utils/programming_utils.hpp
brryan: Nice, changed, thanks

src/phoebus_utils/programming_utils.hpp
brryan: Right now I use this for checking string arguments against a vector of acceptable values (e.g. radiation method):
```c++
  std::vector<std::string> known_methods = {"cooling_function", "moment_m1",                         
                                            "moment_eddington", "monte_carlo", "mocmc"};             
  if (!programming::ContainedInVector(method, known_methods)) {                                      
    std::stringstream msg;                                                                           
    msg << "Radiation method \"" << method << "\" not recognized!";                                  
    PARTHENON_FAIL(msg);                                                                             
  }
  ```
  Maybe this could automatically work with a set or map just by doing something like
  ```c++
template <template <typename> class V, typename T>                                                   
bool ContainedInVector(T &item, V<T> &container) {
```
-- I tried the above but it didn't work so I'm moving on for now

src/phoebus_utils/root_find.hpp
brryan: This isn't even fully implemented yet so I'm adding a `PARTHENON_FAIL` to the top of the function

src/radiation/closure.hpp
brryan: Oh this must have gotten lost when I went through a fairly involved merge conflict. Re-added, thanks.

src/radiation/closure_mocmc.hpp
brryan: Changed to `ClosureStatus::fail`. Maybe a virtual base class would clean this up a bit but I'm not sure it's urgent right now.

src/radiation/kd_grid.hpp
brryan: @jdolence

src/radiation/mocmc.cpp
brryan: Since we'll eventually go to packing with swarms presumably I'd rather not remove the templating boilerplate now only to reintroduce it later

src/radiation/mocmc.cpp
brryan: Yes. Now to add `par_scan` to parthenon (I left in an implementation that works on my CPU, but commented out for now)

src/radiation/mocmc.cpp
brryan: NVM `GetFourVelocity` is already in relativity utils :)

brryan: I guess I'm starting to worry about increasing the complexity of relativity utils any further... to avoid the situation where you end up calling a function that calculates gcov and then calls another subfunction that also calculates gcov for example.

src/radiation/mocmc.cpp
Yurlungur: I mean yeah write the code with the packed API, but if you just change the type and remove the template it's like a 2 line change and you catch the error at compile time instead of run time? idk nbd either way

src/phoebus_utils/programming_utils.hpp
Yurlungur: OOOH I see. Okay this is actually pretty useful then.

Your example should be this:
```C++
template <template <typename> class V, typename T, typename... Args>                                                   
bool ContainedInVector(T &item, V<T, Args...> &container) {
```
because the STL containers often have a bunch of hidden default template parameters.

Doesn't matter though. I only argued for a different type because I didn't realize what you were doing here.

src/phoebus_utils/interpolation.hpp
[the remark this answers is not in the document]
Yurlungur: yes I think that's right. I think there may be some unification possible but we shouldn't think too hard about it for now.

src/radiation/mocmc.cpp
Yurlungur: yeah that's a concern we should be aware of. Not sure what the solution is except try to design things with the ability to pass in other needed quantities. idk

src/phoebus_utils/programming_utils.hpp
brryan: oops, removed, thanks

src/phoebus_utils/programming_utils.hpp
brryan: Yeah something about the `std::find` syntax makes me want to avoid it. Between that and `std::vector<bool>` being broken, I think it is enough justification for us to have our own `pvector` class in here that mostly just wraps `std::vector` but adds, like, a `.contains(T x)` method 🤷

src/radiation/mocmc.cpp
brryan: Yeah, or like receive a struct that checks at compile time what geometric and physical quantities have already been calculated and uses `constexpr if` to calculate anything needed but not yet known. Sounds pretty complicated.

src/phoebus_utils/programming_utils.hpp
Yurlungur: I think we have `std::vector` all over the place in `parthenon`, so this may not be practical, but `std::set` is ordered, unique, and has a `contains` method.

src/radiation/mocmc.cpp
Yurlungur: kk nbd for now

src/radiation/mocmc.cpp
Yurlungur: That sounds way too complicated, lol. For now I think we punt. But something to be aware of.

src/radiation/mocmc.cpp
Yurlungur: :+1:

src/phoebus_utils/programming_utils.hpp
brryan: I just looked -- isn't `std::set::contains` a C++20 feature? https://en.cppreference.com/w/cpp/container/set/contains

Yurlungur: Yes but you can just use `std::set::count` and treat it as a boolean.

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,947 @@
...
+                opac_d.AngleAveragedAbsorptionCoefficient(
+                    v(iblock, pdens, k, j, i), v(iblock, pT, k, j, i),
+                    v(iblock, pye, k, j, i), dev_species[ispec], nusamp(nu_bins - 1)) *
+                v(iblock, Inu0(ispec, nu_bins - 1), k, j, i) * nusamp(nu_bins - 1);
+            Itot -= 0.5 *
+                    (v(iblock, Inu0(ispec, 0), k, j, i) * nusamp(0) +
+                     v(iblock, Inu0(ispec, nu_bins - 1), k, j, i) * nusamp(nu_bins - 1));
+            kappaJ = robust::ratio(kappaJ, Itot);
+            kappaH = kappaJ;
+            kappaJ *= (1. - scattering_fraction);
```
lroberts36, line 683: We should probably add at least a TODO here to remove the `scattering_fraction` correction. I could see it causing some issue in the future when we have physical scattering in.

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,947 @@
...
+
+            Real tauJ = alpha * dt * kappaJ;
+            Real tauH = alpha * dt * kappaH;
+
+            // Store kappaH for asymptotic fluxes
+            v(iblock, idx_kappaH(ispec), k, j, i) = kappaH;
+
+            auto status = c.LinearSourceUpdate(Estar, cov_Fstar, con_tilPi, JBB, tauJ,
+                                               tauH, &dE, &cov_dF);
+
```
lroberts36, line 693: I had initially thought we would have separate tasks for the MOCMC opacity calculation/MOCMC update and the source update for the moments and have fields for kappaJ and kappaH.  The advantage I see there is no duplicated code and maybe more possibility to interleave work, at the cost of more storage.

src/radiation/moments.cpp
```diff
@@ -882,20 +988,18 @@ TaskStatus MomentCalculateOpacities(T *rc) {
           /// opacity so that we can
```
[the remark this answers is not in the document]
lroberts36: Yes, I think I just forgot to remove that TODO when I added those grid variables. 

My original thinking was that the integrated opacities required for the moment source update could be filled by a task that uses the MOCMC samples (i.e. `MOCMCCalculateOpacities`) or by a simple task like this one (`MomentCalculateOpacities`) that just assumes some gray opacity based on the properties of the moments (or by doing something dumb and assuming all neutrinos are 10 MeV).

lroberts36 (commented):
Nice work @brryan! This looks pretty good to me as is, I had only a few comments none of which are major.

inputs/radiation_advection.pin
```diff
@@ -42,8 +42,13 @@ nghost = 4
 nx1         = 512      # Number of zones in X1-direction
 x1min       = 0        # minimum value of X1
 x1max       = 1        # maximum value of X1
-ix1_bc      = user     # Inner-X1 boundary condition flag
-ox1_bc      = user     # Outer-X1 boundary condition flag
+#ix1_bc      = user     # Inner-X1 boundary condition flag
+#ox1_bc      = user     # Outer-X1 boundary condition flag
+# TODO(BRR) Currently an issue with asymptotic preserving transport in samples
+# in pure scattering problems that makes this problem misbehave with outflow
+# boundaries
```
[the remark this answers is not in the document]
brryan: Suppose we have an infinite scattering opacity, no absorption opacity. In MOCMC the moment energy density will just sit there, but the particles will still fly around at `c`. The particle intensities are updated towards the angle-averaged intensity of all particles in that zone, so at late time every particle in the simulation will have equal intensities i.e. the initial radiation energy density will be completely smeared out.

src/pgen/radiation_advection.cpp
```diff
@@ -39,32 +41,65 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
   auto &coords = pmb->coords;
   auto pmesh = pmb->pmy_mesh;
   int ndim = pmesh->ndim;
 
-  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::entire);
-  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::entire);
-  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::entire);
+  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
```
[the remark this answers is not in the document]
brryan: In general I think we want to move to `IndexDomain::interior` for initial conditions, because there's no guarantee that the function providing the initial conditions will be consistent with the chosen boundary conditions. So either way we should be overwriting the ghost cells before the first simulation timestep, but I think since the values from the initial conditions functions in the ghost cells should never be used by the code, it is more clear to use `IndexDomain::interior` in `pgen`.

There was an issue before where boundary conditions weren't always being called before the first timestep so we had to write initial conditions to `IndexDomain::entire`, but again this was inconsistent with our boundaries.

src/radiation/moments.cpp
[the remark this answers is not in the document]
brryan: OK that makes sense, thanks.

I agree it would be nice to just have some task that updates opacities, either via some mean opacity or MOCMC samples. For at least some methods for MOCMC fluid sourcing, though, (in particular if we iterate on the fluid temperature and so need updated opacities each iteration) I had trouble reconciling opacities-via-separate-task with updating opacities each iteration. Admittedly we're not doing that right now, but we might want to later. One could still always call the opacities task prior to any interactions and then update opacities further inside the interaction loop if necessary... Maybe something to be refined later.

src/phoebus_utils/unit_conversions.hpp
brryan: Oh nice, I didn't know you could call other constructors in initializer lists like in that second option. I like that a lot, I just implemented it, thanks :+1:

scripts/plot_thermalization.py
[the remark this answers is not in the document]
brryan: Now plots the phoebus solution along with analytic

src/phoebus_utils/variables.hpp
```diff
@@ -54,8 +54,17 @@ constexpr char dJ[] = "r.i.dJ";
...
 constexpr char kappaH[] = "r.i.kappaH";
 constexpr char JBB[] = "r.i.JBB";
+constexpr char tilPi[] = "r.i.tilPi";
+constexpr char kappaH_mean[] = "r.i.kappaH_mean";
 } // namespace radmoment_internal
 
+namespace mocmc_internal {
+constexpr char dnsamp[] = "mocmc.i.dnsamp";
+constexpr char Inu0[] = "mocmc.i.inu0";
+constexpr char Inu1[] = "mocmc.i.inu0";
```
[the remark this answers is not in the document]
brryan: Oops, no, thanks

src/radiation/mocmc.cpp
brryan: Agreed, `TODO` added. Adding a gray scattering opacity to `singularity-opac` is near the top of my todo list so hopefully I get to this soon.

src/radiation/mocmc.cpp
brryan: I mentioned above that I wondered how to fold a separate opacity calculation task together with an iterative update task -- maybe the solution is to have a function that calculates opacities, a task that wraps that function, and then a fluid source task that can call the opacity function additional times if necessary?

src/phoebus_boundaries/phoebus_boundaries.cpp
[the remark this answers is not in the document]
Yurlungur: I am working on an improvement for this in my branch. We should coordinate. I could spin my thing off and merge it ahead of time maybe.

src/phoebus_utils/phoebus_interpolation.hpp
```diff
@@ -11,8 +11,8 @@
 // distribute copies to the public, perform publicly and display
 // publicly, and to permit others to do so.
 
-#ifndef PHOEBUS_UTILS_PHOEBUS_INTERPOLATION_HPP_
-#define PHOEBUS_UTILS_PHOEBUS_INTERPOLATION_HPP_
+#ifndef PHOEBUS_UTILS_PHOEBUS_INTERPOLATION_
+#define PHOEBUS_UTILS_PHOEBUS_INTERPOLATION_
```
Yurlungur, line 16: You dropped this: `HPP_`

src/radiation/moments.cpp
```diff
@@ -102,20 +111,26 @@ TaskStatus MomentCon2PrimImpl(T *rc) {
...
-        c.GetCovTilPiFromCon(E, covF, xi, phi, &conTilPi);
-        if (STORE_GUESS) {
-          v(b, iXi(ispec), k, j, i) = xi;
-          v(b, iPhi(ispec), k, j, i) = phi;
+        if (programming::is_specialization_of<CLOSURE, ClosureMOCMC>::value) {
+          SPACELOOP2(ii, jj) { conTilPi(ii, jj) = v(b, iTilPi(ispec, ii, jj), k, j, i); }
+        } else {
+          Real xi = 0.0;
+          Real phi = pi;
+          if (STORE_GUESS) {
```
Yurlungur, line 119: Why not always store guess?

Yurlungur (approved):
I don't fully understand everything that's in here, and would appreciate a code walkthrough at some point, so I defer to @lroberts36 and @jdolence for most technical questions. But a few nitpicks below.

Yurlungur:
Overall really exciting to have this in

src/radiation/moments.cpp
brryan: This is really @lroberts36's moment code -- he may have a use case

Yurlungur:
...after tests pass. Looks like still seeing test failures?

src/radiation/moments.cpp
lroberts36: There is no reason not to store the guess except that fields are not allocated for storage when we are not doing M1 (the only place a guess is required).

lroberts36 (approved):
This looks good to me!

src/radiation/mocmc.cpp
```diff
@@ -0,0 +1,948 @@
...
+  // Calculate array of starting index for each zone to compute particles
+  ParArrayND<int> starting_index("Starting index", nx_k, nx_j, nx_i);
+  auto starting_index_h = starting_index.GetHostMirror();
+  auto dN = rc->Get(im::dnsamp).data;
+  auto dN_h = dN.GetHostMirrorAndCopy();
+  int index = 0;
+  for (int k = 0; k < nx_k; k++) {
+    for (int j = 0; j < nx_j; j++) {
+      for (int i = 0; i < nx_i; i++) {
+        starting_index(k, j, i) = index;
```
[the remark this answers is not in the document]
brryan: Yes definitely. I should test this PR on GPU in detail, at the moment though it seems like the radiation diffusion test is failing even with M1 (in a way that `cuda-gdb` doesn't backtrace, annoyingly). I'll look into it.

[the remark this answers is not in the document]
brryan: Actually the failure I attributed to radiation was actually due to flux/src diags for fluid still running for an inactive fluid: https://github.com/lanl/phoebus/pull/115

PR #84, opened 2022-02-25
Make robust c2p faster

jdolence:
Adds some new root finders and improves the logic of con2prim_robust so that it only overwrites the conserved variables when necessary.  Fixes a few other bugs and introduces a couple new pgens.  One that sanity checks p2c and c2p by going back and forth a bunch of times and the other the relativistic MHD rotor.


cmake/Fluid.cmake
```diff
@@ -18,7 +18,7 @@ else()
   set(USE_VALENCIA 0 CACHE BOOL "Covariant formulation of GRMHD")
 endif()
 
-option(PHOEBUS_C2P_ROBUST_FLOORS "Use full floors for c2p robust" OFF)
+option(PHOEBUS_C2P_ROBUST_FLOORS "Use full floors for c2p robust" ON)
```
brryan, line 21: Do we want this `ON` by default? My understanding from your recent disk runs is that for torus calculations we will always want this `OFF`

src/fluid/con2prim_robust.hpp
```diff
@@ -398,26 +394,42 @@ class ConToPrim {
...
+    }
 
     for (int i = 0; i < sig_hi - sig_lo + 1; i++) {
       v(sig_lo + i) = sig[i];
     }
 
     num_nans = std::isnan(v(crho)) + std::isnan(v(cmom_lo)) + std::isnan(v(ceng));
 
-    if (num_nans > 0) {
+    if (num_nans > 0 || res.used_gamma_max()) {
```
brryan, line 432: We may want to revisit this later to decide if `used_gamma_max()` really warrants a stencil fixup... but it's kind of an extreme failure that algorithms like this hate and the non-stencil solution is to basically shave off the maximum anyway so this may be fine. Just mentioning it.

brryan (approved):
A few small queries, but if you don't mind just glancing at them this LGTM and I'll then merge

Glad we have the rotor problem in here

src/pgen/torus.cpp
```diff
@@ -256,17 +263,19 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
+          // JMM: HARM/bhlight divides by gdet, not gamdet.
+          // This means the HARM primitives are smaller than the Phoebus primitives by
+          // a factor of alpha.
+          const Real gamdet = geom.DetGamma(CellLocation::Cent, k, j, i);
           v(iblo, k, j, i) = -(A(j, i) - A(j + 1, i) + A(j, i + 1) - A(j + 1, i + 1)) /
-                             (2.0 * coords.Dx(X2DIR, k, j, i) * gdet);
+                             (2.0 * coords.Dx(X2DIR, k, j, i) * gamdet);
           v(iblo + 1, k, j, i) = (A(j, i) + A(j + 1, i) - A(j, i + 1) - A(j + 1, i + 1)) /
-                                 (2.0 * coords.Dx(X1DIR, k, j, i) * gdet);
+                                 (2.0 * coords.Dx(X1DIR, k, j, i) * gamdet);
```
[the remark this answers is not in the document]
brryan: oh ok thanks

cmake/Fluid.cmake
jdolence: That's not the sense I have.  My sense was actually the opposite.  It seems to actually work quite well with it `ON`.  We clearly need to do some more quantitative analysis as you suggested offline, but I'd vote for this to be `ON` by default, for now at least.

cmake/Fluid.cmake
```diff
@@ -32,3 +32,10 @@ if(PHOEBUS_FLUX_AND_SRC_DIAGS)
 else()
   set(SET_FLUX_SRC_DIAGS 0 CACHE BOOL "Skip these diagnostics")
 endif()
+
+option(PHOEBUS_CON2PRIM_STATISTICS "Gather and report a histogram of iteration counts" OFF)
```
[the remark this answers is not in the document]
jdolence: yeah good idea.

src/fluid/con2prim_robust.hpp
jdolence: yeah, I think it's still unclear exactly when we want to say the inversion failed.  I imagine we'll figure this out as we start to run more production models.

cmake/Fluid.cmake
brryan: Oh OK I thought with the full floors that we were getting a big halo around the torus initially. If that's not true that's great.

PR #89, opened 2022-03-16
Clang Format

Yurlungur:
I think it's about time we formatted the repository, as it's getting pretty messy. This PR does the following:
- I implement a simple script in `bash/scripts/format.sh`, which queries git for every cpp and hpp file in git history, excluding submodules, and calls clang-format on all of them. Note a file must be committed to git for this script to format it.
- I add a github workflow which checks for formatting on PRs. It will fail if another run of clang-format doesn't match.
- I format the repo.

What do you all think? Are you happy with adding the workflow? Are you happy with the massive format commit? Does the script make sense to you? If you're happy, I think this should be a quick merge.


Yurlungur:
There's currently a github actions outage, so the CI isn't triggering. Hopefully that will recover soon.

scripts/bash/format.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+for f in $(git grep --cached -Il res -- :/*.hpp :/*.cpp); do
+    clang-format -i $(git rev-parse --show-toplevel)/${f}
```
brryan, line 4: Should we force a specific `clang-format` here, e.g. `clang-format-10`?

brryan (approved):
Thanks for doing this, it's long overdue. Probably best to wait for github actions (soon, I assume?) to come back before merging.

scripts/bash/format.sh
Yurlungur: I thought about that. I don't really want to... because it's possible clang-format will be stable on some version and it's annoying to force the user to pin to a specific version. But if we start seeing incompatible formats on the CI and on people's commits maybe we need to revisit that. What do you think?

brryan: I agree it's annoying but historically I think `clang-format` has not been totally stable across versions. Maybe our code is un-fancy enough that this isn't an issue though. What about comparing the major version number from `clang-format --version` with some preferred version, like
```bash
CLANG_FORMAT_PREFERRED_VERSION=10
# Regex magic
echo "Warning! clang-format version does not match preferred major version $CLANG_FORMAT_PREFERRED_VERSION!"
```

.github/workflows/formatting.yml
```diff
@@ -0,0 +1,24 @@
...
+      runs-on: ubuntu-latest
+
+      steps:      
+        - name: Checkout code
+          uses: actions/checkout@v2
+        - name: Set system to non-interactive mode
+          run: export DEBIAN_FRONTEND=noninteractive
+        - name: install dependencies
+          run: |
+            sudo apt-get install -y --force-yes -qq git clang-format
```
brryan, line 21: It would probably good at least to have the CI stick with a particular `clang-format`... could this be accomplished by changing this line to
```
sudo apt-get install -y --force-yes -qq git clang-format-10
```
?

.github/workflows/formatting.yml
```diff
@@ -0,0 +1,24 @@
...
+        - name: check formatting
+          continue-on-error: true
+          run: find . -regex '.*\.\(cpp\|hpp\)' | xargs clang-format -style=file -i && git diff --exit-code --ignore-submodules
```
brryan, line 24: This would be `clang-format-10` as well

scripts/bash/format.sh
Yurlungur: Sure I'll do that.

Yurlungur: ok looks like my VM is also on 10 so let's pin to that I guess.

Yurlungur:
@brryan I think I've addressed all comments. I now:
- Pin to clang-format-12 (checked that this is actually installable on both my linux boxes and available on Darwin. Hopefully this means it works for you too.)
- Modify the `format.sh` script so that it warns if clang-format is the wrong version. You can also point it to a specific binary with a command line variable. It also now errors out if there's no clang-format available.
- Reformatted
- Fixed the issues that were causing the tests to fail... I believe having to do with the `test_utils.hpp` header.

PR #90, opened 2022-03-23
Some tweaks of the formatting machinery

Yurlungur:
The CI format was passing no matter what. I removed the line making this happen. I also tweak the `format.sh` script for Ben's use-case. In particular:
- Script always outputs the clang-format binary it found
- Script always outputs the clang-format version. It's up to the user to discern whether or not the version is correct. I didn't want to have to parse every possible format string.


Yurlungur:
@brryan are you happy for me to merge this? or is something still wrong?

brryan (approved):
Happy with this now, thanks @Yurlungur

PR #91, opened 2022-03-24
Add timestep controls to the monopole gravity solver

Yurlungur:
## PR Summary

As we discussed, we currently use lagged variables (i.e., from one subcycle behind) to source the monopole solver. We decided that a totally un-lagged procedure is, at the moment, not desirable. These changes are intended to lend a bit more confidence to our current approach.

I add two features that should add confidence:
1. I add a time step control that demands the time step be shorter than `dtfac * alpha/(dalpha/dt)`, which provides an approximate characteristic time scale for metric evolution. There's no stability criterion here... But presumably if we are taking time steps larger than the metric evolution time scale, then lagging our variables is certainly wrong.
2. I add a new task to the task list for the monopole solver that tries to estimate how good first-order operator splitting is. It compares a finite differences (dalpha/dt) to the analytic one computed by the solver and checks that these two rates are relatively close. It optionally spits out a warning if they are not.

Interested to hear what people think.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/inputs/tov.pin
+++ b/inputs/tov.pin
@@ -135,10 +135,10 @@ run_n_times = -1
 # dtfac is for time step control. 0 < dtfac <= 1.
 # time step will be <= dtfac*(time rate of change of lapse)
 dtfac = 0.9
-# dtwarn tries to estimate if first-order operator split is sufficiently accurate.
+# warn_on_dt tries to estimate if first-order operator split is sufficiently accurate.
 # Compares (dalpha/dt) computed analytically with numerical difference
 # between two subcycles. Warns if difference is larger than dtwarn_eps
-dtwarn = false
+warn_on_dt = false
 dtwarn_eps = 1e-5
 
 <tov>
--- a/inputs/tov.pin
+++ b/inputs/tov.pin
@@ -132,6 +132,14 @@ force_static = false
 # Runs the first n subcycles. Then freezes. Source terms are not disabled.
 # -1 means it's always run.
 run_n_times = -1
+# dtfac is for time step control. 0 < dtfac <= 1.
+# time step will be <= dtfac*(time rate of change of lapse)
+dtfac = 0.9
+# dtwarn tries to estimate if first-order operator split is sufficiently accurate.
+# Compares (dalpha/dt) computed analytically with numerical difference
+# between two subcycles. Warns if difference is larger than dtwarn_eps
+dtwarn = false
+dtwarn_eps = 1e-5
 
 <tov>
 enabled = true
--- a/src/monopole_gr/monopole_gr.cpp
+++ b/src/monopole_gr/monopole_gr.cpp
@@ -528,7 +528,8 @@ TaskStatus CheckRateOfChange(StateDescriptor *pkg, Real dt) {
     if (err > dtwarn_eps) {
       std::cerr
           << "\tWarning! (alpha^i+1 - alpha^i)/dt is very different from dalpha/dt.\n"
-          << "\t\tTolerance set to " << dtwarn_eps << std::endl;
+          << "\t\tTolerance set to " << dtwarn_eps << " but value is " << err << "."
+          << std::endl;
     }
   }
   return TaskStatus::complete;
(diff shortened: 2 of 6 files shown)


Yurlungur:
Now tested and seems to behave itself.

src/monopole_gr/monopole_gr.cpp
```diff
@@ -57,6 +61,17 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   int nth_call = 0;
   params.Add("nth_call", nth_call, true); // mutable
 
+  // Time step control. Equivalent to CFL
+  // Empirical. controls
+  Real dtfac = pin->GetOrAddReal("monopole_gr", "dtfac", 0.9);
+  params.Add("dtfac", dtfac);
+
+  // Warn if the change in lapse is too different from dalpha/dt
+  bool dtwarn = pin->GetOrAddBoolean("monopole_gr", "warn_on_dt", false);
```
lroberts36, line 70: A little confusing that `dtwarn` is the name in the code and `warn_on_dt` is the name in the input file.

src/monopole_gr/monopole_gr.cpp
```diff
@@ -424,15 +491,50 @@ TaskStatus SpacetimeToDevice(StateDescriptor *pkg) {
...
+    auto gradients = params.Get<Gradients_t>("gradients");
+    auto dtwarn_eps = params.Get<Real>("dtwarn_eps");
+
+    Real err;
+    parthenon::par_reduce(
+        parthenon::loop_pattern_flatrange_tag, "monopole_gr check evolution time scales",
+        parthenon::DevExecSpace(), 0, npoints - 1,
+        KOKKOS_LAMBDA(const int i, Real &lmax_err) {
+          Real diff_alpha = (alpha(i) - alpha_last(i)) / dt;
+          Real diff_dalpha = std::abs(diff_alpha - gradients(Gradients::DALPHADT, i));
```
lroberts36, line 524: Should this be relative or absolute difference? Lapse could be small where the action is, right?

src/monopole_gr/monopole_gr.hpp
```diff
@@ -78,6 +78,25 @@
...
+
+  da/dr = a (4 - 4 a^2 + 3 a^2 (r K^r_r)^2 + 32 pi rhob)/(8 r)
+
+  d K^r_r = (8 pi Srb/a - 3 r K^r_r)/(r^2)
+
+  and the equation for alpha as
+
+  a^2 r alpha ((3/2) (K^r_r)^2 + 4 pi ((rho/(a r^2) + S)) +
+  ((1/a)(da/dr)r-2)dalpha/dr
+
```
lroberts36, line 99: I did not check these equations.

Yurlungur: That's fine. They are unused for the moment. I just wanted it documented somewhere if/when we revisit. It's also in the Phoebus doc.

inputs/tov.pin
```diff
@@ -132,6 +132,14 @@ force_static = false
 # Runs the first n subcycles. Then freezes. Source terms are not disabled.
 # -1 means it's always run.
 run_n_times = -1
+# dtfac is for time step control. 0 < dtfac <= 1.
```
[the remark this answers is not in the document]
Yurlungur: It's not enforced. I will add a check. Agreed, there's no reason it couldn't work for larger values... but let's enforce it for now and see if that causes problems down the line.

src/monopole_gr/monopole_gr.cpp
Yurlungur: Hmm. I'm not sure. You're right alpha could be small where the action is. I don't want to renormalize by dalpha/dt, because that would risk dividing by zero when things are well-behaved. But we could divide by alpha to try to capture that scale.

lroberts36: I think your suggestion of absolute differences of logarithmic derivatives is good.

Yurlungur:
Thanks, @lroberts36 ! I addressed all your comments and will go ahead and merge now.

Yurlungur:
or as soon as tests pass

PR #94, opened 2022-04-01
Enable time-dependent cached metric

Yurlungur:
## PR Summary

This PR is in preparation for two separate needs. The first, simpler, need is to support time-dependent analytic coordinate systems and the caching thereof. Examples, might be co-rotating coordinates or metrics that are evolving in a pre-prescribed way. The second is the update on the grid of metrics that change via numerics---by evolving or by elliptic solve. This will support, for example, the monopole GR metric in Cartesian coordinates, rather than just spherical. There's still a bit more work to support the latter case, but this PR provides several changes needed to make that happen and tests those changes.

The changes I make here are:
1. I modify the cached coordinate system so that it can (with run-time selection) store time derivatives of the metric. I also update the geometry machinery to support re-setting and filling the cache every cycle. 
2. I add some internal machinery to support time-dependent metrics, and for the caching machinery to automatically infer if a metric is time-dependent. This is supported by adding three new parameters to the caching machinery: `time_dependent`, `time`, and `reset_coords`. The former is assumed false, unless set to true, and is used in the task list and by the cache to infer whether or not a geometry needs to be updated. A geometry should set it to true if it wants to be updated every cycle. The latter is similar, but determines whether, upon an update, the `coords` fields in the output get updated every cycle for visualization purposes. This is to support, e.g., co-rotating coordinate systems. The middle param, `time`, need only be added for metrics that have an analytic prescription that depends on time. If it's present, it will be updated by the driver so that they geometry package knows what physical time a given cycle is.

I also add a simple test to check all this machinery works. In particular, I construct a one-zone test with a cosmological (homogeneous and isotropic) metric. I pre-prescribe the scale factor to have constant time derivative and then solve for the density as a function of time, which will decay as a power law as determined by the second [Friedmann equation](https://en.wikipedia.org/wiki/Friedmann_equations). The solution as a function of time converges to first-order with dt (because the metric is lagged) but agreement is nevertheless excellent. See the image bleow:

image

Interested to hear your thoughts @jdolence @brryan @lroberts36 . If you all like it, I will move on to using this machinery to enable on-grid finite differences for the Cartesian monopole solver, which should simplify the code and be pretty clean and performant.

## PR Checklist

- [x] Explain what you did.


Yurlungur:
Oh one other thing: I would like to add the cosmological metric as a regression test, but it seems kind of silly to make a gold file. @brryan do you have thoughts? The gold file method is perhaps the path of least resistance... so maybe that's the path we should go down?

src/geometry/cached_system.hpp
```diff
@@ -456,93 +479,98 @@ void SetCachedCoordinateSystem(MeshBlockData<Real> *rc) {
...
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "SetGeometry::Set Cached data, Corn", DevExecSpace(), 0,
+      pack.GetDim(5) - 1, kbs, kbe, jb.s, jb.e + 1, ib.s, ib.e + 1,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+        lamb(b, k, j, i, CellLocation::Corn);
       });
 
   // don't let other kernels launch until geometry is set
-  pmb->exec_space.fence();
+  Kokkos::fence();
```
jdolence, line 573: do we actually need this fence?

src/phoebus_driver.cpp
```diff
@@ -148,6 +149,17 @@ TaskCollection PhoebusDriver::RungeKuttaStage(const int stage) {
     pmono_vol_red = monopole->MutableParam<MonoVolRed_t>("volumes_reducer");
   }
 
+  // Hack to update time for time-dependent geometries
+  auto geometry = pmesh->packages.Get("geometry");
+  auto &geom_params = geometry->AllParams();
+  // tm = SimTime field in EvolutionDriver. See parthenon/src/driver/driver.hpp
+  if (geom_params.hasKey("time")) {
+    const Real tstage = tm.time + robust::ratio((stage - 1) * tm.dt,
```
jdolence, line 157: is this right?  for our default form of RK2, the prime state is a t+dt not t+dt/2.

jdolence (changes requested):
just a couple quick things to address.  overall this came out really clean.  nice job.

src/phoebus_driver.cpp
Yurlungur: Oh hmm we're using Heun's method. Maybe this is the right thing, under full generality...
```C++
const Real tstage = tm.time + (stage > 1 ? (integrator->beta[stage-2]) : 0)*tm.dt;
```
so for stage 1 it's just the current time. Then for subsequent stages, it's the time + `beta*dt` What do you think?

src/geometry/cached_system.hpp
Yurlungur: I could be convinced to drop it... My concern was a race condition where the geometry variables are accessed before they're filled. i.e., if you call `pgen` before the geometry vars are done.

Yurlungur:
@dbbowen this is potentially of interest to you.

src/phoebus_driver.cpp
jdolence: yeah I think this is right

Yurlungur:
In the process of formatting this, I discovered a bug in the formatting script (and a missing copyright in a file). Both have been fixed.

Yurlungur:
I also added a regression test for the FLRW metric.

PR #100, opened 2022-04-08
fix B-field normalization in torus pgen

Yurlungur:
…Bsq calc in tmunu and torus

## PR Summary

What it says on the can. The issue is that the density is already normalized when the vector potential is computed. Dividing by an additional factor of `rho_rmax` shrinks `A` and, thanks to the `-0.2` in the definition, kills `A`, and thus `B`, at the edges of the torus.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/fluid/tmunu.hpp
+++ b/src/fluid/tmunu.hpp
@@ -122,7 +122,7 @@ class StressEnergyTensorCon {
       b[l] = iW * (b_(l, std::forward<Args>(args)...) + alpha * b[0] * u[l]);
     }
 
-    bsq = (Bsq + alpha * alpha * b[0] * b[0]) * iW * iW;
+    bsq = Bsq * iW * iW + Bdotv * Bdotv;
   }
 
   Pack pack_;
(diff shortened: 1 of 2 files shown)


Yurlungur:
There was also a missing factor of `alpha`^2 in the calculation of `bsq` in the normalization routine that sets beta, meaning beta was incorrectly computed.

jdolence:
good catch.  trying to understand the substantive changes here.  From my read, the only thing actually changing is that we were erroneously dividing by `rho_rmax`, which this removes.  The rest of the edits don't look like they actually change anything.  Is that right?

Yurlungur:
There was also a missing alpha^2 in the calculation of bsq used when computing beta but I think it probably didn't matter because b0 should be zero in the initial conditions.

jdolence:
> There was also a missing alpha^2 in the calculation of bsq used when computing beta but I think it probably didn't matter because b0 should be zero in the initial conditions.

I don't think the alpha^2 was missing because that wasn't actually b[0] it was W * Bdotv.  Anyway, I guess you're right that it won't matter for this setup, but it would matter later for other setups (like toroidal fields).

Yurlungur:
Oh you're right. Well in that case all I did there was clarify things a little for own sanity. 

The rho_rmax was the only substantive change.

PR #101, opened 2022-04-12
reduce geometry memory footprint

Yurlungur:
## PR Summary

This is based on a discussion with @jdolence about what geometry objects we actually need to access and from where. The geometry cache, especially the metric derivatives, take up a lot of memory. I am currently storing all geometry objects at all cell centers, faces, and nodes. However, we only really need metric derivatives at cell centers (for source terms) and the nodes are totally untouched.

This PR removes node-centered variables entirely, except for "physical coordinate locations," which is useful for 3D visualization tools like visit. It also removes metric derivatives from cell faces. This reduces the memory footprint of the caching machinery substantially... by a bit more than a factor of 2.

The price, of course, is that metric derivatives---which are needed to compute the christoffel symbols---are no longer available at faces and nodes. I don't believe we need them there, but I could be wrong. In particular, @brryan do we need the christoffel symbols in the ghost zones? If so, face- or node-centered data is probably useful for interpolation. If not, though, we can probably get away with these savings.

Another price is to be paid in the API for the geometry. I now have the code die (in debug mode) if you try to access node-centered data where it is not available. Is this an acceptable price?

I'm interested in folk's thoughts. If people like this, then we should run a torus, and the snake linear modes test, with caching on to make sure I didn't break anything.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/geometry/cached_system.hpp
+++ b/src/geometry/cached_system.hpp
@@ -115,9 +115,9 @@ class Cached {
     }
     // These variables only exist at cell centers
     PARTHENON_DEBUG_REQUIRE(imap["g.c.dalpha"].second >= 0, "variable exists");
-    PARTHENON_DEBUG_REQUIRE(
-        imap["g.c.dalpha"].second - imap["g.c.dalpha"].first + 1 == nvar_deriv_,
-        "variable shape correct");
+    PARTHENON_DEBUG_REQUIRE(imap["g.c.dalpha"].second - imap["g.c.dalpha"].first + 1 ==
+                                nvar_deriv_,
+                            "variable shape correct");
     idx_[CellLocation::Cent].dalpha = imap["g.c.dalpha"].first;
     PARTHENON_DEBUG_REQUIRE(imap["g.c.dg"].second >= 0, "variable exists");
     idx_[CellLocation::Cent].dg = imap["g.c.dg"].first;
--- a/src/geometry/cached_system.hpp
+++ b/src/geometry/cached_system.hpp
@@ -396,6 +396,11 @@ void InitializeCachedCoordinateSystem(ParameterInput *pin, StateDescriptor *geom
   bool axisymmetric = pin->GetOrAddBoolean("geometry", "axisymmetric", false);
   params.Add("axisymmetric", axisymmetric);
 
+  // This tells the goemetry infrastructure not to call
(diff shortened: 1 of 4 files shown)


src/geometry/cached_system.hpp
```diff
@@ -409,6 +428,18 @@ void InitializeCachedCoordinateSystem(ParameterInput *pin, StateDescriptor *geom
...
     geometry->AddField("g.c." + var_names[i], m);
   }
+  // Do metric derivative and lapse derivative separately, as
+  // they're only available on cell centers
+  for (int i = 0; i < cent_only_sizes.size(); ++i) {
+    if (axisymmetric) {
+      shape = {dims.nx1, dims.nx2, 1, cent_only_sizes[i]};
+      m = Metadata(flags_o, shape);
+    } else {
+      shape = {cent_only_sizes[i]};
```
brryan, line 438: Should this not be `shape = {dims.nx1, dims.nx2, dims.nx3, cent_only_sizes[i]};`?

brryan (commented):
LGTM apart from one comment. Apart from that and checking a 2D GRMHD torus really quick, I'm happy with this, nice optimization.

src/geometry/cached_system.hpp
Yurlungur: No, it can be just `cent_only_sizes[i]`. If you only pass in 1 number, and tell Parthenon that your variable is cell-centered, it infers the other dimension.

brryan: Ah right, thanks :+1:

brryan (approved):

Yurlungur:
Running on the torus revealed a mistake on my end with the `Coords` array in `CachedSystem`. I do need to set in `CachedSystem` as `SetGeometryDefault` was using the values in the cache. 

It also revealed two long-standing bugs:
1. The `Coords` function in `ModifiedSystem` was incorrectly returning the coordinate transformation vector, not the Cartesian coordinates for visualization.
2. `SetCachedGeometry` and `SetGeometryDefault` were both filling the `Coords` array, duplicating both code and work. To fix this this, I make `SetCachedGeometry` *call* `SetGeometryDefault` and have `SetGeometryBlock` only call `SetGeometryDefault` if the `Coords` array actually needs to be set.

PR #103, opened 2022-04-15
Change default normalization of B-field in torus to the HARM-style normalization

Yurlungur:
## PR Summary

What it says on the can. This was required to get the torus to look more like the one we're all used to. There's a switch in the problem generator to use one or the other. I also made several quality of life changes:
1. I changed the problem modifier to set up the grid exactly as bhlight does by default. This moves the inner boundary further into the horizon. The problem modifier now reports what it changed x1min and x1max to, as well as what the inner radius is.
2. I wrote some routines to calculate b^2 in `relativity_utils.hpp`
3. I added a routine to calculate the beta used for normalization via the minimum value or the old HARM style. The problem generator now reports both values.
4. I changed the default inner and outer radii of the torus in the input deck to match the standard values for a 2D torus in HARM/bhlight.
5. I moved all the utility functions in `torus.cpp` to the bottom of the file, with some prototypes at the top. I find this easier to read.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/pgen/torus.cpp
+++ b/src/pgen/torus.cpp
@@ -11,82 +11,41 @@
 // distribute copies to the public, perform publicly and display
 // publicly, and to permit others to do so.
 
+// C/C++ includes
+#include <cstdio>
+#include <limits>
+
+// Kokkos
 #include "Kokkos_Random.hpp"
+
+// Parthenon
+#include <globals.hpp>
+
+// Phoebus
 #include "fluid/con2prim_robust.hpp"
 #include "geometry/boyer_lindquist.hpp"
 #include "geometry/mckinney_gammie_ryan.hpp"
 #include "pgen/pgen.hpp"
 #include "phoebus_utils/reduction.hpp"
+#include "phoebus_utils/relativity_utils.hpp"
 #include "phoebus_utils/robust.hpp"
 
 typedef Kokkos::Random_XorShift64_Pool<> RNGPool;
 
 namespace torus {
 
-Real lfish_calc(Real r, Real a) {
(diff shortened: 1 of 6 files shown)


Yurlungur:
Because I was sick of waiting, I also modified the movie scripts to render the frames in parallel.

src/pgen/torus.cpp
```diff
@@ -295,21 +257,143 @@ void ProblemModifier(ParameterInput *pin) {
...
+            gcov[3][3] * ucon[3] * ucon[3] +
+            2. * (gcov[1][2] * ucon[1] * ucon[2] + gcov[1][3] * ucon[1] * ucon[3] +
+                  gcov[2][3] * ucon[2] * ucon[3]);
+  Real discr = BB * BB - 4. * AA * CC;
+  if (discr < 0) printf("discr = %g   %g %g %g\n", discr, AA, BB, CC);
+  PARTHENON_REQUIRE(discr >= 0, "discr < 0");
+  return (-BB - std::sqrt(discr)) / (2. * AA);
+}
+
+// TODO(JMM): Should this be elsewhere in Phoebus?
```
brryan, line 391: I think this is the best spot for this right now, since it calculates global quantities (I'm not sure that `reduction::Min` and `reduction::Max` are going to behave outside of problem initilalization)

brryan (approved):
LGTM, nice cleanups too

src/pgen/torus.cpp
Yurlungur: :+1:

PR #104, opened 2022-04-15
Radiative torus

brryan:
## PR Summary

Add radiation as an option to the torus problem. Either zero initial radiation or equipartition of Fishbone solution pressure between fluid and radiation inside the initial torus.

## PR Checklist

- [x] Explain what you did.
- [x] Loop radiation floors and ceilings through parameters interface, and code
- [x] Remove Eddington assumption from various radiation calls
- [x] Get `GetFlatIdx` to return default values when variable not included in pack
- [x] Rebuild driver to agree with `main`
- [x] Loop over `nspec` for radiation sources
- [x] Loop over `nspec` during `FixFluxes`
- [x] Clean up `plot_snap2d.py`
- [x] Clean up `torus_rad.pin`
- [ ] Replace `scattering_fraction` with separate scattering opacity call
- [ ] Add `kappa_max` to radiation ceilings, propagate through radiation fluid source code
- [x] Add method for choosing between radiation source term methods
- [x] Replace `NumRadiationTypes` with `MaxNumRadiationTypes`
- [x] Fix the uncountably infinite number of radiation-related warnings I've introduced in the past
- [ ] Figure out why compile times are so much slower with `nvcc` in this PR


src/microphysics/opac_phoebus/opac_phoebus.cpp
```diff
@@ -36,6 +36,12 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 
   bool do_rad = pin->GetBoolean("physics", "rad");
   if (!do_rad) {
+    // TODO(BRR) Allow for copy constructor in opacity variant so we don't need to always provide
+    // a default opacity
```
Yurlungur, line 40: This should work now

src/pgen/torus.cpp
```diff
@@ -29,11 +29,43 @@
...
 
 namespace torus {
 
+using pc = parthenon::constants::PhysicalConstants<parthenon::constants::CGS>;
+
+using namespace radiation;
+using namespace singularity;
+using namespace singularity::neutrinos;
+
+class Residual {
```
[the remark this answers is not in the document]
brryan: `GasRadTemperatureResidual`

src/radiation/moments.cpp
```diff
@@ -15,14 +15,75 @@
...
+          mean_opacity_.RosselandMeanAbsorptionCoefficient(rho_, T, Ye_, species_[ispec]);
+
+      const Real JBB = opacity_.EnergyDensityFromTemperature(T, species_[ispec]);
+
+      dJ_tot += (J0_[ispec] + dtau_ * kappa * JBB) / (1. + dtau_ * kappa) - J0_[ispec];
+    }
+
+    const Real ug1 = rho_ * eos_.InternalEnergyFromDensityTemperature(rho_, T, lambda);
+
+    return ((ug1 - ug0_) + (dJ_tot)) / (ug0_ + J0_tot);
```
brryan, line 70: @lroberts36 Here is the implicit-in-fluid-temperature residual. I solve `ug(T^n+1) - ug(T^n) = ur^n+1 - ur^n` in the fluid frame. Then after I get `T^n+1`, I use `LinearSourceUpdate` to get `dE` and `dF` and update the conserved radiation and fluid variables as usual.

src/radiation/moments.cpp
```diff
@@ -804,29 +1124,94 @@ TaskStatus MomentFluidSource(T *rc, Real dt, bool update_fluid) {
...
+            v(iblock, ceng, k, j, i) += alpha * sdetgam * dE;
+#endif
             v(iblock, cmom_lo + 0, k, j, i) -= sdetgam * cov_dF(0);
             v(iblock, cmom_lo + 1, k, j, i) -= sdetgam * cov_dF(1);
             v(iblock, cmom_lo + 2, k, j, i) -= sdetgam * cov_dF(2);
           }
-        }
+        } // for ispec
+
+        /* for (int ispec = 0; ispec < num_species; ++ispec) {
```
brryan, line 1142: I'll go back and make the source update algorithm a runtime option, probably a templated `FluidSourceImpl` function like we do elsewhere.

src/radiation/moments.cpp
lroberts36: I like this, it is a good solution.

brryan (commented):
@lroberts36 @Yurlungur This is finally ready for re-review.

src/fluid/con2prim_robust.hpp
```diff
@@ -429,7 +435,8 @@ class ConToPrim {
 
     num_nans = std::isnan(v(crho)) + std::isnan(v(cmom_lo)) + std::isnan(v(ceng));
 
-    if (num_nans > 0 || res.used_gamma_max()) {
+    // if (num_nans > 0 || res.used_gamma_max()) {
```
brryan, line 434: I think the gamma max ceiling can be safely caught subsequently by `ApplyFloors` which will limit us to `gamma_max`

tst/regression/mocmc_diffusion.gold
```diff
@@ -1,76 +1,25 @@
...
+-1.188232745820429572e-17
+-4.992172249544863943e-20
+3.600838140944280810e-20
+3.223026185159611409e-18
+3.207937073915787390e-18
+-1.678974299604244268e-18
+4.709490397007393027e-19
+-3.950108603179772350e-18
+-2.606358899989489921e-19
+1.986847194014894714e-19
```
brryan, line 25: These gold files shrunk because we aren't always outputting 3 rad species of data any more

tst/regression/regression_test.py
```diff
@@ -208,10 +208,13 @@ def gold_comparison(variables, input_file, modified_inputs={},
     modify_input(key, modified_inputs[key], TEMPORARY_INPUT_FILE)
 
   # Run test problem
+  preamble = []
+  if use_mpiexec:
+      preamble = preamble + ['mpiexec', '-n', '1']
```
brryan, line 213: `mpiexec` is optional now.

tst/regression/linear_modes.gold
```diff
@@ -1,1232 +1,1232 @@
...
--1.400635785566348553e-06
-1.648279431176801968e-07
-1.613534133241710520e-06
-1.504597441069569458e-06
-1.028065202879326984e-08
--1.507090455995950591e-06
--1.655940491521968514e-06
--3.579361683397125794e-07
-1.261762431799203453e-06
+1.000005632783612652e+00
```
brryan, line 1: Im not sure why this needed an upgold but still converges at 2nd order when I checked with `scripts/python/linmode.py`

brryan:
OK actually lets hold off on this for just a second, my git commits weren't quite synced up between machines and I have to fix a few things. Hopefully today.

brryan:
> OK actually lets hold off on this for just a second, my git commits weren't quite synced up between machines and I have to fix a few things. Hopefully today.

OK my confusion was just that I had a compile-time parameter set differently between darwin and my local machine. Turned that (whether to lower the floors used by c2p robust relative to the "real" floors to avoid atmosphere zones at the floors constantly failing) into a runtime parameter to avoid this issue in the future. This should be ready to go again.

src/pgen/torus.cpp
```diff
@@ -197,6 +266,69 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
+                           v(iprs, k, j, i);
+
+        Real ucon[4] = {0};
+        Real vpcon[3] = {v(ivlo, k, j, i), v(ivlo + 1, k, j, i), v(ivlo + 2, k, j, i)};
+        Real gcov[4][4] = {0};
+        geom.SpacetimeMetric(CellLocation::Cent, k, j, i, gcov);
+        GetFourVelocity(vpcon, geom, CellLocation::Cent, k, j, i, ucon);
+
+        // Radiation
+        if (do_rad) {
```
Yurlungur, line 280: This is not necessarily what we're always going to want. Should it specify equipartition or something?

src/pgen/torus.cpp
```diff
@@ -268,10 +405,20 @@ void ProblemModifier(ParameterInput *pin) {
...
+  // Set Cv properly for radiation
+  const bool do_rad = pin->GetOrAddBoolean("physics", "rad", false);
+  if (do_rad) {
+    const std::string eos_type = pin->GetString("eos", "type");
+    if (eos_type == "IdealGas") {
+      const Real Gamma = pin->GetReal("eos", "Gamma");
+      const Real cv = (Gamma - 1.) * pc::kb / pc::mp;
+      pin->SetReal("eos", "Cv", cv);
+    }
+  }
```
Yurlungur, line 418: also not guaranteed to be what we want

src/phoebus_boundaries/phoebus_boundaries.cpp
```diff
@@ -132,29 +138,64 @@ void OutflowInnerX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
...
-            vcon[0] = beta[0] / alpha;
-            Real vsq = 0.;
-            SPACELOOP2(ii, jj) { vsq += gammacov[ii][jj] * vcon[ii] * vcon[jj]; }
-            W = 1. / sqrt(1. - vsq);
-
-            SPACELOOP(ii) { q(pv_lo + ii, k, j, i) = W * vcon[ii]; }
-          }
         });
+
+    // TODO(BRR) Is this always unnecessary if BC inside BH?
```
Yurlungur, line 143: We can always get this back with the magic of version control. I say we delete it.

src/radiation/closure.hpp
```diff
@@ -139,21 +139,31 @@ class ClosureEdd {
...
+  // TODO(BRR) use gamma max ceiling (may mess with rootfind)
+  // v2 = std::min<Real>(v2, 0.9999);
   W = 1 / std::sqrt(1 - v2);
+  if (std::isinf(W)) {
+    printf("Nonsense W! vin = %e %e %e\n", con_v_in(0), con_v_in(1), con_v_in(2));
+    printf("v = %e %e %e\n", con_v(0), con_v(1), con_v(2));
+    printf("cov_v: %e %e %e\n", cov_v(0), cov_v(1), cov_v(2));
+    printf("v2 = %e\n", v2);
+    printf("W = %e\n", W);
+  }
```
Yurlungur, line 159: either print rank and to stderr or fail after all this?

src/phoebus_driver.cpp
```diff
@@ -398,46 +367,7 @@ TaskCollection PhoebusDriver::RungeKuttaStage(const int stage) {
...
-      auto impl_update = tl.AddTask(update, radiation::MOCMCFluidSource<MeshData<Real>>,
-                                    sc1.get(), beta * dt, fluid_active);
-      auto impl_edd =
-          tl.AddTask(impl_update, radiation::MOCMCEddington<MeshData<Real>>, sc1.get());
-      update = impl_update | update;
-    } else if (rad_moments_active) {
-      auto impl_update = tl.AddTask(update, radiation::MomentFluidSource<MeshData<Real>>,
-                                    sc1.get(), beta * dt, fluid_active);
-      update = impl_update | update;
+      sync_region_6.AddRegionalDependencies(reg_dep_id++, ib, interp_to_monopole);
```
Yurlungur, line 370: it's hard to see what's going on here from the diff... it's ideal to minimize the number of regions, since that minimizes fences. Are you doing that?

src/phoebus_driver.cpp
```diff
@@ -483,51 +413,323 @@ TaskCollection PhoebusDriver::RungeKuttaStage(const int stage) {
...
-    auto recv = tl.AddTask(
-        update, parthenon::cell_centered_bvars::ReceiveBoundBufs<nonlocal>, sc1);
-    auto set = tl.AddTask(recv, parthenon::cell_centered_bvars::SetBounds<nonlocal>, sc1);
+    if (pmesh->multilevel) {
+      tl.AddTask(set_nonlocal | set_local,
+                 parthenon::cell_centered_refinement::RestrictPhysicalBounds, mc1.get());
+    }
+  }
+
+  // TODO(BRR) loop this in thoughtfully!
```
Yurlungur, line 514: reminder

src/phoebus_driver.cpp
```diff
@@ -605,6 +807,7 @@ parthenon::Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
   packages.Add(Microphysics::EOS::Initialize(pin.get()));
   packages.Add(Microphysics::Opacity::Initialize(pin.get()));
   packages.Add(Geometry::Initialize(pin.get()));
+  // packages.Add(radiation::Initialize(pin.get()));
```
Yurlungur, line 810: is there no radiation package anymore?

src/radiation/closure_m1.hpp
```diff
@@ -266,10 +264,12 @@ KOKKOS_FUNCTION M1Result ClosureM1<Vec, Tens2, SET>::SolveClosure(Real E, Vec co
...
-template <class Vec, class Tens2, class SET>
-KOKKOS_FUNCTION ClosureStatus ClosureM1<Vec, Tens2, SET>::M1Residuals(
-    const Real E, const Vec cov_F, const Real xi, const Real phi, const Vec con_tilg,
-    const Vec con_tild, Real *fXi, Real *fPhi) {
+template <class SET>
+KOKKOS_FUNCTION ClosureStatus ClosureM1<SET>::M1Residuals(const Real E, const Vec cov_F,
+                                                          const Real xi, const Real phi,
+                                                          const Vec con_tilg,
+                                                          const Vec con_tild, Real *fXi,
+                                                          Real *fPhi) {
```
Yurlungur, line 272: is this just clang format?

src/radiation/mocmc.cpp
```diff
@@ -277,22 +274,19 @@ TaskStatus MOCMCSampleBoundaries(T *rc) {
...
   }
 
-  /// TODO: (LFR) Fix this junk
-  RadiationType dev_species[3] = {species[0], species[1], species[2]};
-
   auto swarm_d = swarm->GetDeviceContext();
 
   auto ix1_bc = rad->Param<MOCMCBoundaries>("ix1_bc");
-  auto ox1_bc = rad->Param<MOCMCBoundaries>("ox1_bc");
+  // auto ox1_bc = rad->Param<MOCMCBoundaries>("ox1_bc");
```
Yurlungur, line 280: clean this up?

src/radiation/moments.cpp
```diff
@@ -514,31 +556,92 @@ TaskStatus CalculateFluxesImpl(T *rc) {
           face = CellLocation::Face3;
           break;
         }
+        Real X[4];
+        X[1] = (face == CellLocation::Face1 ? coords.x1f(k, j, i) : coords.x1v(k, j, i));
+        X[2] = (face == CellLocation::Face2 ? coords.x2f(k, j, i) : coords.x2v(k, j, i));
+        X[3] = (face == CellLocation::Face3 ? coords.x3f(k, j, i) : coords.x3v(k, j, i));
+
+        Real W_ceiling, garbage;
```
Yurlungur, line 564: lol `garbage`. Maybe make the final argument optional?

src/radiation/moments.cpp
```diff
@@ -608,14 +715,15 @@ TaskStatus CalculateFluxesImpl(T *rc) {
 
           // Calculate the numerical flux using LLF
           v.flux(idir_in, idx_Ef(ispec), k, j, i) =
-              0.5 * sdetgam * (conFl(idir) + conFr(idir) + speed * (El - Er));
+              0.5 * sdetgam * (conFl(idir) + conFr(idir) + sigspeed * (El - Er));
```
Yurlungur, line 718: good namechange :+1:

src/radiation/moments_source.cpp
```diff
@@ -0,0 +1,801 @@
+// © 2022. Triad National Security, LLC. All rights reserved.
```
Yurlungur, line 1: I like that this is it's own file now.

src/radiation/moments_source.cpp
```diff
@@ -0,0 +1,801 @@
...
+    const Real E = U_rad[0] / sdetgam_;
+    Vec cov_F{U_rad[1] / sdetgam_, U_rad[2] / sdetgam_, U_rad[3] / sdetgam_};
+    Vec cov_H;
+    Real W = 0.;
+    SPACELOOP2(ii, jj) { W += (*gcov_)[ii + 1][jj + 1] * P_mhd[ii + 1] * P_mhd[jj + 1]; }
+    W = std::sqrt(1. + W);
+    Vec con_v{P_mhd[1] / W, P_mhd[2] / W, P_mhd[3] / W};
+    // Use gamma_max from code?
+    if (W > 100) {
+      printf("W = %e! [%i %i %i]\n", W, k_, j_, i_);
```
Yurlungur, line 97: print to stderr or do debug warn or debug fail?

src/radiation/moments_source.cpp
```diff
@@ -0,0 +1,801 @@
...
+  const auto scattering_fraction = rad->Param<Real>("scattering_fraction");
+
+  const auto src_solver = rad->Param<SourceSolver>("src_solver");
+  const auto src_use_oned_backup = rad->Param<bool>("src_use_oned_backup");
+  const auto src_rootfind_eps = rad->Param<Real>("src_rootfind_eps");
+  const auto src_rootfind_tol = rad->Param<Real>("src_rootfind_tol");
+  const auto src_rootfind_maxiter = rad->Param<int>("src_rootfind_maxiter");
+  const auto oned_fixup_strategy = rad->Param<OneDFixupStrategy>("oned_fixup_strategy");
+
+  parthenon::par_for(
```
Yurlungur, line 295: This kernel is an absolute beast, and difficult to parse. Anyway to break it up or at least have some device functions to call?

src/radiation/radiation.cpp
```diff
@@ -304,27 +305,115 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+    const std::string src_solver_name =
+        pin->GetOrAddString("radiation", "src_solver", "fourd");
+    SourceSolver src_solver;
+    if (src_solver_name == "zerod") {
+      src_solver = SourceSolver::zerod;
+    } else if (src_solver_name == "oned") {
+      src_solver = SourceSolver::oned;
+    } else if (src_solver_name == "fourd") {
+      src_solver = SourceSolver::fourd;
+    } else {
```
Yurlungur, line 363: will we always need all these?

src/radiation/radiation.cpp
```diff
@@ -304,27 +305,115 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+    //    std::string solver = pin->GetOrAddString("radiation", "riemann", "hll");
+    //    riemann::solver rs = riemann::solver::HLL;
+    //    if (solver == "llf") {
+    //      rs = riemann::solver::LLF;
+    //    } else if (solver == "hll") {
+    //      rs = riemann::solver::HLL;
+    //    } else {
+    //      PARTHENON_THROW("Invalid Riemann Solver option. Choose from [llf, hll]");
+    //    }
+    //    params.Add("RiemannSolver", rs);
```
Yurlungur, line 352: delete?

src/radiation/radiation.cpp
```diff
@@ -304,27 +305,115 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+    } else if (oned_fixup_strategy_str == "ignore_all") {
+      oned_fixup_strategy = OneDFixupStrategy::ignore_all;
+    } else {
+      PARTHENON_FAIL("radiation/oned_fixup_strategy has an invalid entry!");
+    }
+    params.Add("oned_fixup_strategy", oned_fixup_strategy);
+
     int ndim = 3;
     // if (pin->GetInteger("parthenon/mesh", "nx3") > 1) ndim = 3;
     // else if (pin->GetInteger("parthenon/mesh", "nx2") > 1) ndim = 2;
```
Yurlungur, line 398: I guess we don't need this? delete?

src/radiation/radiation.cpp
```diff
@@ -354,39 +443,63 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
     // Add variables for source functions
     Metadata mSourceVar = Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy},
-                                   std::vector<int>{NumRadiationTypes});
+                                   std::vector<int>{num_species});
     physics->AddField(i::kappaJ, mSourceVar);
     physics->AddField(i::kappaH, mSourceVar);
     physics->AddField(i::JBB, mSourceVar);
 
+    // this fail flag should really be an enum or something
+    // but parthenon doesn't yet support that kind of thing
```
Yurlungur, line 457: it's fine we can be evil

Yurlungur (approved):
So this is a huge diff. I'm not going to claim I absorbed it all. But I think this mostly seems reasonable. Some comments below.

tst/regression/regression_test.py
brryan: We could do something like that but then all of a sudden we'd need per-machine configurations. Maybe something to revisit in phoebus if we have time for firelink in the future.

src/fixup/fixup_c2p.cpp
```diff
@@ -0,0 +1,407 @@
...
+using radiation::Vec;
+using robust::ratio;
+using singularity::RadiationType;
+using singularity::neutrinos::Opacity;
+
+namespace fixup {
+
+template <typename T, class CLOSURE>
+TaskStatus ConservedToPrimitiveFixupImpl(T *rc, T *rc0,
+                                         IndexDomain domain = IndexDomain::interior) {
```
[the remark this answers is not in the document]
brryan: Yes you're right. With the quite robust c2p_robust this is probably less of a concern... Let me try a GRMHD torus with cons var ghost filling. I think we still want to fill physical boundaries with prims, I'm not sure if theres a more elegant way to do this than just by tagging prim + cons with FillGhost (although some "dependent" primitives like `gam1`, `P`, `T` wouldn't need to be FillGhost-ed.

inputs/radiation_advection.pin
```diff
@@ -105,6 +105,7 @@ absorption = true
 do_nu_electron = true
 do_nu_electron_anti = false
 do_nu_heavy = false
+src_solver = oned
```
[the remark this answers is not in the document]
brryan: I'd like to hang onto the 4D solver for now in case there are parameter regimes where it is more accurate. One could imagine using it as a first attempt when `Pr >> Pg` or something. For now `oned` should be the hardcoded default, though.

scripts/python/phoebus_eos.py
```diff
@@ -26,6 +26,9 @@ def T_from_rho_u_Ye(self, rho, u, Ye):
   def u_from_rho_T_Ye(self, rho, T, Ye):
     raise NotImplementedError()
 
+  def P_from_rho_u_Ye(self, rho, u, Ye):
```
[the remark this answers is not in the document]
brryan: Oh neat -- let me open an issue though and do that in a subsequent PR -- #128

scripts/python/phoedf.py
```diff
@@ -21,10 +21,13 @@
 # Phoebus-specific derived class from Parthenon's phdf datafile reader
 class phoedf(phdf.phdf):
 
-  def __init__(self, filename):
+  def __init__(self, filename, geomfile=None):
```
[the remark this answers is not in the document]
brryan: Yeah I'd be interested to see how it works. Though the python pipeline is pretty deeply ingrained in my muscle memory at this point :)

src/fixup/fixup.cpp
```diff
@@ -43,13 +54,34 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   bool enable_floors = pin->GetOrAddBoolean("fixup", "enable_floors", false);
   bool enable_c2p_fixup = pin->GetOrAddBoolean("fixup", "enable_c2p_fixup", false);
   params.Add("enable_c2p_fixup", enable_c2p_fixup);
+  bool enable_source_fixup = pin->GetOrAddBoolean("fixup", "enable_source_fixup", false);
+  params.Add("enable_source_fixup", enable_source_fixup);
   bool enable_ceilings = pin->GetOrAddBoolean("fixup", "enable_ceilings", false);
   params.Add("enable_ceilings", enable_ceilings);
+  bool enable_rad_ceilings = pin->GetOrAddBoolean("fixup", "enable_rad_ceilings", false);
```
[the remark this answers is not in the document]
brryan: Yeah agreed

src/fixup/fixup.cpp
```diff
@@ -43,13 +54,34 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  params.Add("enable_rad_ceilings", enable_rad_ceilings);
   bool report_c2p_fails = pin->GetOrAddBoolean("fixup", "report_c2p_fails", false);
   params.Add("report_c2p_fails", report_c2p_fails);
+  bool report_source_fails = pin->GetOrAddBoolean("fixup", "report_source_fails", false);
+  params.Add("report_source_fails", report_source_fails);
   bool enable_mhd_floors = pin->GetOrAddBoolean("fixup", "enable_mhd_floors", false);
+  bool enable_rad_floors = pin->GetOrAddBoolean("fixup", "enable_rad_floors", false);
```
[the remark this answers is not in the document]
brryan: Yep. Also I switched this to default to `true` if rad physics is enabled else default to `false`

src/fixup/fixup.cpp
```diff
@@ -43,13 +54,34 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+
+  const bool enable_mhd = pin->GetOrAddBoolean("fluid", "mhd", false);
+  const bool enable_rad = pin->GetOrAddBoolean("physics", "rad", false);
```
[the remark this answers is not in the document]
brryan: `mhd` defaults to `false` in `fluid/fluid.cpp` so I'd like to keep the same logic here

src/fixup/fixup.cpp
```diff
@@ -113,38 +150,135 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  if (fluid_c2p_failure_strategy_str == "interpolate") {
+    fluid_c2p_failure_strategy = FAILURE_STRATEGY::interpolate;
+  } else if (fluid_c2p_failure_strategy_str == "interpolate_previous") {
+    fluid_c2p_failure_strategy = FAILURE_STRATEGY::interpolate_previous;
+  } else if (fluid_c2p_failure_strategy_str == "neighbor_minimum") {
+    fluid_c2p_failure_strategy = FAILURE_STRATEGY::neighbor_minimum;
+  } else if (fluid_c2p_failure_strategy_str == "floors") {
+    fluid_c2p_failure_strategy = FAILURE_STRATEGY::floors;
+  }
+  params.Add("fluid_c2p_failure_strategy", fluid_c2p_failure_strategy);
```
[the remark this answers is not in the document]
brryan: `interpolate` averages the updated primitives from good neighbors. `interpolate_previous` uses the primitives from the previous substep for interpolation. `neighbor_minimum` uses updated good primitives but takes the minimum of each primitive over the good neighbors. `floors` sets everything to floors. 

Since this ended up not being necessary I should probably just remove everything except `interpolate` (which falls back to `floors` in the case of no good neighbors) and `floors`.

inputs/radiation_advection.pin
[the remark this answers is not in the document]
Yurlungur: :+1:

src/fixup/fixup.cpp
```diff
@@ -226,259 +379,317 @@ TaskStatus ApplyFloors(T *rc) {
...
-            floor_applied = true;
-            v(b, peng, k, j, i) = bsq / bsqou_max;
-          }
-          if (v(b, peng, k, j, i) / v(b, prho, k, j, i) > uorho_max) {
-            floor_applied = true;
-            v(b, peng, k, j, i) = uorho_max * v(b, prho, k, j, i);
-          }
-        }
+          rho_floor_max = std::max<Real>(rho_floor_max, bsq / bsqorho_max);
+          u_floor_max = std::max<Real>(u_floor_max, bsq / bsqou_max);
```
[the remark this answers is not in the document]
brryan: The cleaner logic is due to @bprather 's KHARMA which I shamelessly lifted :)

src/radiation/radiation.hpp
```diff
@@ -43,12 +43,18 @@ enum class MOCMCRecon { kdgrid };
 
 enum class MOCMCBoundaries { outflow, fixed_temp, periodic };
 
+enum class SourceSolver { zerod, oned, fourd };
+
+enum class OneDFixupStrategy { none, ignore_dJ, ignore_all };
+
+enum class ReconFixupStrategy { none, bounds };
```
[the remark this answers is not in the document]
brryan: I should add comments here. `ReconFixupStrategy::none` means just use the reconstructed variables for radiation fluxes and don't worry about them. `ReconFixupStrategy::bounds` applies the floors/ceilings to the reconstructed variables (both rad variables and reconstructed velocities) before using them to compute fluxes.

src/fixup/fixup_c2p.cpp
```diff
@@ -0,0 +1,456 @@
...
+            v(b, iv, k, j, i) += fail(k, j - 1, i) * v(b, iv, k, j - 1, i) +
+                                 fail(k, j + 1, i) * v(b, iv, k, j + 1, i);
+            if (ndim == 3) {
+              v(b, iv, k, j, i) += fail(k - 1, j, i) * v(b, iv, k - 1, j, i) +
+                                   fail(k + 1, j, i) * v(b, iv, k + 1, j, i);
+            }
+          }
+          return inv_mask_sum * v(b, iv, k, j, i);
+        };
+        auto get_minimum = [&](const int iv) {
```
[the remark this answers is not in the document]
brryan: Not a bad idea. Though this kernel is now called so rarely/never (thankfully) that it may not be worth the time of the refactor at the moment.

src/fixup/fixup_radc2p.cpp
```diff
@@ -0,0 +1,364 @@
...
+using radiation::Tens2;
+using radiation::Vec;
+using robust::ratio;
+using singularity::RadiationType;
+using singularity::neutrinos::Opacity;
+
+namespace fixup {
+
+template <typename T, class CLOSURE>
+TaskStatus RadConservedToPrimitiveFixupImpl(T *rc, T *rc0) {
```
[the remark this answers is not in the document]
brryan: Pattern due to @lroberts36 :+1:

src/fixup/fixup_radc2p.cpp
```diff
@@ -0,0 +1,364 @@
...
+              const Real norm = 1.0 / num_valid;
+              for (int ispec = 0; ispec < nspec; ispec++) {
+                v(b, idx_J(ispec), k, j, i) = fixup(idx_J(ispec), norm);
+                SPACELOOP(ii) {
+                  v(b, idx_H(ispec, ii), k, j, i) = fixup(idx_H(ispec, ii), norm);
+                }
+              }
+            } else {
+              for (int ispec = 0; ispec < nspec; ispec++) {
+                v(b, idx_J(ispec), k, j, i) = 1.e-100;
```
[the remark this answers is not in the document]
brryan: I've had some issues doing that with e.g. checking that variables are greater than `robust::SMALL()` as DBCs. I'm not sure what the right pattern is since this right here is a magic number.

src/fluid/con2prim_robust.hpp
```diff
@@ -53,17 +51,16 @@ class Residual {
...
+        floor_scale_fac_(floor_scale_fac) {
     lambda_[0] = Ye;
     Real garbage = 0.0;
     bounds_.GetFloors(x1_, x2_, x3_, rho_floor_, garbage);
     bounds_.GetCeilings(x1_, x2_, x3_, gam_max_, e_max_);
 
-#if !USE_C2P_ROBUST_FLOORS
-    rho_floor_ = 1.e-20;
-#endif
+    rho_floor_ *= floor_scale_fac_;
```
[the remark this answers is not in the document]
brryan: This is the trick that makes the torus happy -- we allow `c2p_robust` to use floors to never fail, but the floors it uses are much smaller than the real floors we subsequently apply in the normal observer frame in `ApplyFloors`.

src/pgen/rhs_tester.cpp
```diff
@@ -20,10 +20,10 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
   auto &rc = pmb->meshblock_data.Get();
 
   PackIndexMap imap;
-  auto v = rc->PackVariables({fluid_prim::density, fluid_prim::velocity,
-                              fluid_prim::energy, fluid_prim::bfield, fluid_prim::ye,
-                              fluid_prim::pressure, fluid_prim::temperature},
-                             imap);
+  auto v = rc->PackVariables(
```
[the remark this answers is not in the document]
brryan: I don't have plans to use it. @jdolence's call

src/pgen/torus.cpp
brryan: Yeah the description of this PR even says that I let the user choose between thermal initial radiation and no initial radiation lol

src/fixup/fixup.cpp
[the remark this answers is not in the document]
Yurlungur: I would be in favor of removing the unused versions :+1:

src/fluid/fluid.cpp
```diff
@@ -190,7 +206,10 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   // Just want constant primitive fields around to serve as
   // background if we are not evolving the fluid, don't need
   // to do the rest.
-  if (!active) return physics;
+  // TODO(BRR) logic gets very complicated accounting for this in situations where
+  // radiation is active but fluid isn't -- for example we want fluid prims from c2p to
+  // calculate opacities.
+  // if (!active) return physics;
```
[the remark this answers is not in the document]
brryan: I feel medium that the comment should remain, since I think there was some thought put into being able to not allocate fluid cons variables for non-fluid runs. One could maybe work around this even for the 4D radiation source solve by not allowing that version of source solving if the fluid isn't active, but I think that extra software engineering work isn't well-motivated right now. I'll make an issue to think about this later, though I don't imagine many scenarios where we really care about this much memory in this case.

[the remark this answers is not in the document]
brryan: #129

src/fluid/fluid.cpp
```diff
@@ -422,7 +444,12 @@ TaskStatus ConservedToPrimitiveRobust(T *rc, const IndexRange &ib, const IndexRa
   StateDescriptor *pkg = pmb->packages.Get("fluid").get();
   const Real c2p_tol = pkg->Param<Real>("c2p_tol");
   const int c2p_max_iter = pkg->Param<int>("c2p_max_iter");
-  auto invert = con2prim_robust::ConToPrimSetup(rc, bounds, c2p_tol, c2p_max_iter);
+  const Real c2p_floor_scale_fac = pkg->Param<Real>("c2p_floor_scale_fac");
+  const bool c2p_fail_on_floors = pkg->Param<bool>("c2p_fail_on_floors");
+  const bool c2p_fail_on_ceilings = pkg->Param<bool>("c2p_fail_on_ceilings");
+  auto invert = con2prim_robust::ConToPrimSetup(rc, bounds, c2p_tol, c2p_max_iter,
+                                                c2p_floor_scale_fac, c2p_fail_on_floors,
+                                                c2p_fail_on_ceilings);
```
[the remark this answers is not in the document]
brryan: Just following the existing pattern :) Our c2p approach could probably be cleaned up a bit at this point, see new issue #130

src/fluid/fluid.cpp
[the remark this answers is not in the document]
Yurlungur: Yeah I find it hard to believe that the memory footprint here will be important. But up to you.

src/phoebus_boundaries/phoebus_boundaries.cpp
brryan: OK. Yeah I was pleasantly surprised to see that this apparently doesn't matter.

src/radiation/closure.hpp
brryan: We've already got a `PARTHENON_DEBUG_REQUIRE` checking the same thing below, I'll just remove this

src/phoebus_driver.cpp
brryan: I should actually take another pass at this now that the "right" algorithm is a bit more clear which should inform what needs to be communicated when. I'll update here when Ive done that.

I think ultimately this will need to be retooled a bit anyway once all our tasks are available on a per-`MeshData` basis (for example, particles).

src/phoebus_driver.cpp
brryan: I moved it to below `fluid` in the hopes that this would change the order in which per-package `FillDerived` calls were called (we need to update rad after fluid because rad prims are in the fluid frame) but it doesn't seem to. I'll delete the commented out extra line.

src/radiation/closure_m1.hpp
brryan: Yeah, I didn't play any special formatting games here.

src/radiation/mocmc.cpp
brryan: This needs to be properly implemented for all boundaries, I'll leave that for my next pr, `brryan/mocmc_torus` (I added a `TODO(BRR)` here)

src/radiation/moments.cpp
brryan: Just following the style from `con2prim_robust.hpp` :) Optional final argument might be a bit too specific to satisfy principle-of-least-astonishment for `GetFloors`, and also maybe we will add another floor in the future

src/radiation/moments_source.cpp
brryan: I'm just going to remove this for now, if we end up sourcing a very fast fluid we'll just clamp the fluid velocity subsequently (and we aren't really using this 4D solve right now anyway).

src/radiation/moments_source.cpp
brryan: I agree, I held off on doing that though because both 1D and 4D solvers require a lot of data that would have to become function arguments. Though maybe I could make a `struct` they share. I could probably reduce code sharing between the different 1D solvers (`regular` and `ignore_dJ`) a bit at least.

src/radiation/radiation.cpp
brryan: Long term we will at *least* need `oned` and a new `twod` (that solves for both `T` and `Ye`). Let me think about whether it's better to just remove `fourd` for now (and revert later if we need it again) -- it would certainly clean up the code a lot and make implementing `twod` a bit less painful.

src/radiation/radiation.cpp
brryan: Yeah thanks.

src/radiation/radiation.cpp
brryan: Actually I think we can reintroduce this since `ndim` is only used for some appropriately guarded reconstruction variables. Not sure why this was ever defaulted to `ndim=3`.

src/radiation/radiation.cpp
brryan: no choice :raised_hands:

src/radiation/moments_source.cpp
brryan: OK thinking more about this I think the right thing to do is a semi-involved refactor (#131) which I would like to defer for a little while. But I'll definitely get to this by the time I add the `twod` solver. That OK?

tst/regression/linear_modes.gold
```diff
@@ -1,1232 +1,1232 @@
-1.000005628528739532e+00
```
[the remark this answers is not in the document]
Yurlungur: I guess if it converges, it's fine, but it would be good to understand where the diff is coming from

src/radiation/radiation.hpp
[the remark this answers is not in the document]
Yurlungur: ah ok :+1: do we ever *not* want to bound reconstructed variables?

src/fixup/fixup_c2p.cpp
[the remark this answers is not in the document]
Yurlungur: :+1:

src/fixup/fixup_radc2p.cpp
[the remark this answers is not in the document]
Yurlungur: `10*robust::SMALL()`?

src/fluid/con2prim_robust.hpp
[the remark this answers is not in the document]
Yurlungur: interesting. I suppose that's probably problem dependent... kind of gross.

src/pgen/rhs_tester.cpp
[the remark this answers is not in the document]
Yurlungur: I initially wrote the `rhs-tester`, not @jdolence ;) I think it's served it's purpose for now... but it could be a useful test if we ever lose confidence that we're doing things sanely in the fluid sector I suppose.

src/pgen/torus.cpp
Yurlungur: Yeah I mean Cv isn't going to be constant for the vast majority of EOS's we're going to run. I think we in general will need to solve for equal temperature value with a root find.

src/radiation/moments.cpp
Yurlungur: yeah ok

src/radiation/moments_source.cpp
Yurlungur: :+1: works for me

src/fixup/fixup.cpp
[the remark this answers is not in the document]
brryan: Removed :+1:

src/fluid/con2prim_robust.hpp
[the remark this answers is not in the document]
brryan: I think it's only as problem-dependent as the actual floors used in `ApplyFloors` since I think the only restriction on the `c2p_robust` floors for torus stability is that they be much smaller than the `ApplyFloors` floors

src/pgen/linear_modes.cpp
```diff
@@ -222,6 +224,9 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
         // k, j, i)/rho);
         v(itmp, k, j, i) = eos.TemperatureFromDensityInternalEnergy(
             rho, v(ieng, k, j, i) / rho, eos_lambda);
+        v(igm1, k, j, i) = eos.BulkModulusFromDensityTemperature(
+                               v(irho, k, j, i), v(itmp, k, j, i), eos_lambda) /
+                           v(iprs, k, j, i);
```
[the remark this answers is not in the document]
brryan: I added DBCs to `PrimitiveToConserved` that in `DEBUG` mode check that for example `v(gm1, ...) > 0`, which started failing :)

src/fluid/con2prim_robust.hpp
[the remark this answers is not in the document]
Yurlungur: hmm ok.

src/pgen/torus.cpp
brryan: This `torus` pgen already does do a rootfind to set the rad and gas to be at equal temperatures (`GasRadTemperatureResidual` at the top of this class). This segment of code setting `cv` just overrules whatever is in the input parameter file so you don't have to worry about correctly calculating `cv` in cgs units when setting that parameter file (previous phoebus inputs are scale-free MHD problems (or scale-free rad) so `cv` was always just set to `1` or something).

src/phoebus_driver.cpp
brryan: The radiation package still exists and is initialized two lines down, I just forgot to uncomment the original place where I removed it

src/pgen/torus.cpp
Yurlungur: Oh I see... Can we not just have some ability to set the units of `Cv` in the input file?

src/phoebus_driver.cpp
Yurlungur: Ah ok :+1:

tst/regression/linear_modes.gold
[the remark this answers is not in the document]
brryan: Oh you know I bet this is because we're now correctly setting `gm1` in initialization in the linear modes pgen:
```c++
 v(igm1, k, j, i) = eos.BulkModulusFromDensityTemperature(
                               v(irho, k, j, i), v(itmp, k, j, i), eos_lambda) /
                           v(iprs, k, j, i);
                           ```

src/radiation/radiation.hpp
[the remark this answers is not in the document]
brryan: I don't immediately see a reason to not do this but I think we should at least get reconstructed eddington tensors in with MOCMC and see how those work with bounds before requiring this feature to always be on.

src/fixup/fixup_radc2p.cpp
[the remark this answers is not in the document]
brryan: Done :+1:

src/pgen/rhs_tester.cpp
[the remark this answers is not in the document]
brryan: OK lets just leave it in for now then, we can always do a separate hygiene PR later

src/radiation/radiation.hpp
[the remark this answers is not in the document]
Yurlungur: kk :+1:

src/pgen/torus.cpp
brryan: well you would still have to enter some wonky number. I'm not sure what units would turn pc::kb/pc::mp into 1

tst/regression/linear_modes.gold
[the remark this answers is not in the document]
Yurlungur: oooh yeah ok.

src/pgen/torus.cpp
Yurlungur: presumably in units of `mass = kb`. :P 

or alternatively we should have some kind of "scale free" mode?

brryan: I'm not sure what the right solution is now. One could maybe have another option for setting scale-free eos that takes mean molecular weight and adiabatic index, say. I've made an issue so this can be deferred: #133

src/pgen/torus.cpp
brryan: OK no initial radiation should work as well now

brryan:
OK if tests pass I think this is ready to merge if you are happy with it @Yurlungur

src/phoebus_driver.cpp
brryan: Now thoughtfully looped in -- I removed a whole c2p + boundary conditions setting, and TWO rounds of send/recv. Code was very redundant previously.

Yurlungur:
I am happy with it

Yurlungur:
Also great work. Awesome to have GRRMHD working.

brryan:
Thanks for your detailed review @Yurlungur! Much appreciated.

PR #108, opened 2022-04-20
 Add the ability to compute metric derivatives by finite differences on the grid

Yurlungur:
## PR Summary

In my on-going quest to build the geometry machinery up to the ability to do on-grid computations for e.g., `MonopoleCart`, in this PR I introduce the ability to compute metric derivatives via finite differences on the grid when using cached geometry.

I allow a given geometry to specify whether or not it should do this when cached via the `geometry/do_fd_on_grid` `param`. If this `param` is not set, the cached geometry will read it in from the input deck.

The finite differences themselves are computed as centered differences where the cell-centered value is set by the differences across faces. This is vastly simplified by PR #101, which removed the metric derivatives at faces and nodes.

I add a new problem generator to be able to test this machinery, though it is not integrated into automated tests, `pgen/check_cached_geom.hpp` which reports the error between metric derivatives for a cached geometry and an analytic one. This has the added bonus of double checking that memory locations in the cache are set correctly.

There will be one follow-up PR for the geometry infrastructure after this one, where I modify the cached geometry to use interpolation when values between grid points are requested, rather than passing through to the underlying geometry object. That final change will be to support, e.g., particles. With that final change, I think it will be straightforward to set up a `MonopoleCart` geometry which does everything on the grid.

Pinging @jdolence @brryan @lroberts36 

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/inputs/linear_modes.pin
+++ b/inputs/linear_modes.pin
@@ -63,10 +63,10 @@ num_threads = 1         # maximum number of OMP threads
 <phoebus/mesh>
 #bc_vars = primitive
 
-<parthenon/meshblock>
-nx1 = 128
-nx2 = 128
-nx3 = 1
+# <parthenon/meshblock>
+# nx1 = 128
+# nx2 = 128
+# nx3 = 1
 
 <parthenon/refinement0>
 field = c.c.bulk.rho
@@ -91,6 +91,7 @@ vy = 0. # 0.524631
 # vy = 0.81     # shift. Default is 0.
 # vy = 0.617213 # shift. Default is 0.
 # vy = 0      # shift. Default is 0.
+do_fd_on_grid = false
 
 <eos>
 type = IdealGas
--- a/inputs/spherical_shock_tube.pin
+++ b/inputs/spherical_shock_tube.pin
@@ -92,10 +92,10 @@ ox3_bc      = periodic      # Outer-X3 boundary condition flfgag
 
 num_threads = 1         # maximum number of OMP threads
 
-<parthenon/meshblock>
-nx1 = 512
-nx2 = 1
-nx3 = 1
+# <parthenon/meshblock>
+# nx1 = 512
+# nx2 = 1
+# nx3 = 1
 
 <parthenon/refinement0>
 field = c.c.bulk.rho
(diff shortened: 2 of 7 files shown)


Yurlungur:
> I add a new problem generator to be able to test this machinery, though it is not integrated into automated tests, pgen/check_cached_geom.hpp which reports the error between metric derivatives for a cached geometry and an analytic one. This has the added bonus of double checking that memory locations in the cache are set correctly.

Unfortunately this problem generator isn't integrated into the regression tests. I don't know how I'd do that without changing the test infrastructure to be more flexible than just gold files. But it can be run as, e.g.,
```bash
export NX=1024 && ./src/phoebus -i ../inputs/linear_modes.pin phoebus/problem=check_cached_geom geometry/do_fd_on_grid=true parthenon/mesh/nx1=${NX} parthenon/mesh/nx2=${NX}
```
There's no real need for this generator to have its own input deck, as the fluid isn't actually set up. I played this game with snake coordinates, spherical Minkowski, spherical Kerr-Schild, and FMKS, giving me some confidence things are working. Some notes:
- In snake coordinates, spherical Minkowski, and spherical Kerr-Schild, things beautifully converge
- In FMKS, things converge to a point, then stop converging... probably due to the fact that the metric derivative is computed via finite differences "not on the grid" in the first place. Given the success of spherical minkowski and snake, I'm not too bothered by this.

inputs/linear_modes.pin
```diff
@@ -63,10 +63,10 @@ num_threads = 1         # maximum number of OMP threads
...
 #bc_vars = primitive
 
-<parthenon/meshblock>
-nx1 = 128
-nx2 = 128
-nx3 = 1
+# <parthenon/meshblock>
+# nx1 = 128
+# nx2 = 128
+# nx3 = 1
```
lroberts36, line 69: What are these changes for?

lroberts36 (approved):
LGTM.

inputs/linear_modes.pin
Yurlungur: This sets the default to one meshblock which just made it easier for me to do the resolution study.

PR #109, opened 2022-04-26
Provide interpolation for Cached Geometry

Yurlungur:
## PR Summary

This is the final piece needed before I embark on Cartesian Monopole. This introduces the ability in the cached coordinate system to compute values off of cell centers/faces via interpolation, rather than falling back to the underlying analytic formulae. For particles this could be significantly cheaper, and it is necessary when gradients are computed by finite differences on the grid.

I enhanced my `check_cached_geom` problem generator to also check interpolations and checked against spherical Kerr-Schild and FLRW coordinates, and things seem to converge nicely.

One word of warning: coordinate singularities, such as at the poles, appear to cause problems (unsurprisingly)---the method still converges, but the convergence degrade as you move towards the singularity. Caution should be taken when using this machinery near a coordinate singularity.

For now I think we punt this issue, but we may want to do something special near the poles when running in spherical coordinates if we discover problems.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/inputs/spherical_shock_tube.pin
+++ b/inputs/spherical_shock_tube.pin
@@ -103,7 +103,7 @@ method = derivative_order_1
 max_level = 3
 
 <geometry>
-axisymmetry = false
+axisymmetric = false
 do_fd_on_grid = false
 
 <eos>
--- a/src/geometry/analytic_system.hpp
+++ b/src/geometry/analytic_system.hpp
@@ -169,6 +169,10 @@ class Analytic {
     return system_.Lapse(X0, X1, X2, X3);
   }
   KOKKOS_INLINE_FUNCTION
+  Real Lapse(int b, Real X0, Real X1, Real X2, Real X3) const {
+    return system_.Lapse(X0, X1, X2, X3);
+  }
+  KOKKOS_INLINE_FUNCTION
   Real Lapse(CellLocation loc, int k, int j, int i) const {
     Real X1, X2, X3;
     indexer_.GetX(loc, k, j, i, X1, X2, X3);
@@ -183,6 +187,10 @@ class Analytic {
 
   KOKKOS_INLINE_FUNCTION
   void ContravariantShift(Real X0, Real X1, Real X2, Real X3, Real beta[NDSPACE]) const {
+    return system_.ContravariantShift(X0, X1, X2, X3, beta);
+  }
+  KOKKOS_INLINE_FUNCTION
+  void ContravariantShift(int b, Real X0, Real X1, Real X2, Real X3, Real beta[NDSPACE]) const {
     return system_.ContravariantShift(X0, X1, X2, X3, beta);
   }
   KOKKOS_INLINE_FUNCTION
@@ -205,6 +213,10 @@ class Analytic {
     return system_.Metric(X0, X1, X2, X3, gamma);
   }
   KOKKOS_INLINE_FUNCTION
+  void Metric(int b, Real X0, Real X1, Real X2, Real X3, Real gamma[NDSPACE][NDSPACE]) const {
+    return system_.Metric(X0, X1, X2, X3, gamma);
+  }
+  KOKKOS_INLINE_FUNCTION
   void Metric(CellLocation loc, int k, int j, int i, Real gamma[NDSPACE][NDSPACE]) const {
     Real X1, X2, X3;
     indexer_.GetX(loc, k, j, i, X1, X2, X3);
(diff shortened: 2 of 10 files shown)


inputs/spherical_shock_tube.pin
```diff
@@ -103,7 +103,7 @@ method = derivative_order_1
 max_level = 3
 
 <geometry>
-axisymmetry = false
+axisymmetric = false
```
Yurlungur, line 106: I got this variable name wrong in my last PR

src/phoebus_utils/grid_utils.hpp
```diff
@@ -0,0 +1,58 @@
...
+  return coord.x1v(i);
+}
+template <>
+KOKKOS_FORCEINLINE_FUNCTION Real GetXv<X2DIR>(const int i, const Coordinates_t &coord) {
+  return coord.x2v(i);
+}
+template <>
+KOKKOS_FORCEINLINE_FUNCTION Real GetXv<X3DIR>(const int i, const Coordinates_t &coord) {
+  return coord.x3v(i);
+}
```
Yurlungur, line 41: I use these in interpolation. Kind of annoying this functionality isn't in `parthenon`. Maybe it should be in a future change?

Yurlungur:
@jdolence @brryan @lroberts36 with this in the code I can do `MonopoleCart`. Ben I especially want your eyes on it because it mucks with the part of the geometry machinery that particles touch.

Yurlungur:
Tests failing because I'm getting an include failure on the CI I'm not getting on my desktop. Will investigate further tomorrow.

Yurlungur:
Phew. Tests pass. Ready for review @jdolence @brryan @lroberts36

src/geometry/cached_system.hpp
```diff
@@ -304,7 +387,7 @@ class Cached {
     const int offset = !time_dependent_;
     SPACETIMELOOP2(mu, nu) {
       const int flat =
-          nvar_deriv_ * Utils::Flatten2(mu, nu, NDFULL) + idx_[loc].dg - offset;
+          nvar_deriv_ * Utils::Flatten2(mu, nu, NDFULL) + idx_[cent_].dg - offset;
```
lroberts36, line 390: What is the origin of this change? It looks like `loc` is now not being used. Should this always return the cell centered value for the metric derivative?

lroberts36 (commented):
This looks like a reasonable approach to me. The only real comment/question I have is about the cell centered location being hardcoded in the metric derivative.

src/geometry/analytic_system.hpp
```diff
@@ -169,6 +169,10 @@ class Analytic {
     return system_.Lapse(X0, X1, X2, X3);
   }
   KOKKOS_INLINE_FUNCTION
+  Real Lapse(int b, Real X0, Real X1, Real X2, Real X3) const {
+    return system_.Lapse(X0, X1, X2, X3);
+  }
```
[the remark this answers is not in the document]
Yurlungur: Good idea. I'll add that.

src/phoebus_utils/grid_utils.hpp
Yurlungur: lol fair enough. In Parthenon, I was imagining something like
```C++
coord.x1v(dir, i);
```
which wouldn't be too different than what's already there. But it's fine here too.

src/phoebus_utils/phoebus_interpolation.hpp
```diff
@@ -0,0 +1,143 @@
...
+                                w[0][1] * p(b, v, ix[2], ix[1] + 1, ix[0] + 1))) +
+          w[2][1] * (w[1][0] * (w[0][0] * p(b, v, ix[2] + 1, ix[1], ix[0]) +
+                                w[0][1] * p(b, v, ix[2] + 1, ix[1], ix[0] + 1)) +
+                     w[1][1] * (w[0][0] * p(b, v, ix[2] + 1, ix[1] + 1, ix[0]) +
+                                w[0][1] * p(b, v, ix[2] + 1, ix[1] + 1, ix[0] + 1))));
+}
+
+/*
+ * Trilinear interpolation on a variable or meshblock pack
+ * PARAM[IN] - b - Meshblock index
```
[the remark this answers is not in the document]
Yurlungur: Yep comment is wrong. Good catch, thanks.

src/geometry/cached_system.hpp
Yurlungur: This is a mistake. I will change it back to `loc`.

src/geometry/analytic_system.hpp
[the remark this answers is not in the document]
Yurlungur: Actually... for analytic systems, there's no requirement `b==0`, because the underlying system presumably knows where it is. An analytic system might be called interchangeably with a cached system, and calls with `b != 0` should work in both cases.

src/geometry/cached_system.hpp
Yurlungur: Resolved.

Yurlungur:
@lroberts36 thanks for the comments. I think they should all be addressed now.

lroberts36 (approved):
LGTM

PR #110, opened 2022-04-28
MonopoleCart rewrite

Yurlungur:
## PR Summary

This PR finally rewrites the Cartesian monopole coordinate system tied to the monopole GR solver. I do the following:
1. Extend `ModifiedSystem` to support geometries with a time-dependent component. The coordinate transformation is assumed to be time-independent, but the geometry itself is allowed to evolve. This basically entails computing the `d/dt` components of the metric gradients and transforming them into the new coordinate system as appropriate, assuming zero cross-talk between the space and time components of the metric. Note this approach **only** works because the coordinate transform is time-independent.
2. Write a `SphericalToCartesian` transformation object, which I use in the `monopole_gr` interpolation from the 3d grid to the 1d grid to transform relevant quantities like the Jacobian.
3. Define the Cartesian monopole geometry as `using MonopoleCart=Modified<MonopoleSph, SphericalToCartesian>`
4. Wrap that in the `Analytic` and `Cached` geometry modifier templates
5. Profit.

I still need to test basically all of this, but it compiles and I think it's ready for feedback from y'all if you want to take a look. @brryan @jdolence @lroberts36 

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/cmake/Geometry.cmake
+++ b/cmake/Geometry.cmake
@@ -1,4 +1,4 @@
-# © 2021. Triad National Security, LLC. All rights reserved.  This
+# © 2021-2022. Triad National Security, LLC. All rights reserved.  This
 # program was produced under U.S. Government contract 89233218CNA000001
 # for Los Alamos National Laboratory (LANL), which is operated by Triad
 # National Security, LLC for the U.S.  Department of Energy/National
--- a/cmake/Geometry.cmake
+++ b/cmake/Geometry.cmake
@@ -31,28 +31,11 @@ set(PHOEBUS_ANALYTIC_GEOMETRIES
     "MonopoleCart"
     )
 
-# List geometries that should not be cached here
-set(PHOEBUS_GEOMETRY_NO_CACHE
-  # "MonopoleSph"
-  )
-
-# List geometries that MUST be cached here
-set(PHOEBUS_GEOMETRY_MUST_CACHE
-  # "MonopoleCart"
-  )
-
 if(PHOEBUS_ANALYTIC_GEOMETRY)
   if (PHOEBUS_GEOMETRY IN_LIST PHOEBUS_ANALYTIC_GEOMETRIES)
     set(GEOMETRY_MESH "Analytic<${PHOEBUS_GEOMETRY}, IndexerMesh>")
     set(GEOMETRY_MESH_BLOCK "Analytic<${PHOEBUS_GEOMETRY}, IndexerMeshBlock>")
-    if (PHOEBUS_GEOMETRY IN_LIST PHOEBUS_GEOMETRY_MUST_CACHE)
-      message(WARNING "The selected geometry must be cached. Enabling cache.")
-      set(PHOEBUS_CACHE_GEOMETRY ON CACHE BOOL "Force geometry to be cached" FORCE)
-    endif()
     if (PHOEBUS_CACHE_GEOMETRY)
-      if (PHOEBUS_GEOMETRY IN_LIST PHOEBUS_GEOMETRY_NO_CACHE)
-	message(FATAL_ERROR "The selected geometry does not support caching.")
-      endif()
       set(GEOMETRY_MESH "CachedOverMesh<${GEOMETRY_MESH}>")
       set(GEOMETRY_MESH_BLOCK "CachedOverMeshBlock<${GEOMETRY_MESH_BLOCK}>")
     endif()
--- a/cmake/Geometry.cmake
+++ b/cmake/Geometry.cmake
@@ -16,6 +16,7 @@ option(PHOEBUS_CACHE_GEOMETRY "Cache geometry. Only available for some geometrie
 # Default
 set(PHOEBUS_GEOMETRY "Minkowski" CACHE STRING "The metric used by Phoebus")
 
+# List analytic geometries here.
 set(PHOEBUS_ANALYTIC_GEOMETRIES
     "Minkowski"
     "BoostedMinkowski"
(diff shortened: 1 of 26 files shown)


cmake/Geometry.cmake
```diff
@@ -29,19 +30,28 @@ set(PHOEBUS_ANALYTIC_GEOMETRIES
     "MonopoleSph"
     "MonopoleCart"
     )
+
+# List geometries that should not be cached here
 set(PHOEBUS_GEOMETRY_NO_CACHE
```
brryan, line 35: Is geometries that shouldn't be cached only for geometries currently under development, or should some geometries *never* be cached? Same question for `MUST_CACHE`.

src/geometry/modified_system.hpp
```diff
@@ -184,10 +193,38 @@ class Modified {
   void MetricDerivative(Real X0, Real X1, Real X2, Real X3,
                         Real dg[NDFULL][NDFULL][NDFULL]) const {
     Utils::SetMetricGradientByFD(*this, dx_, X0, X1, X2, X3, dg);
+    // If time-dependent, assume coordinate transformation is time
+    // indepedent, pull dg_{mu nu}/dt from underlying system, and
+    // transform mu and nu
+    if (time_dependent_) {
```
brryan, line 199: These `if (time_dependent_)` checks may lead to a lot of unnecessary `if`s deep in compute loops, no? Since the geometry is set at compile time anyway, should this be an `#ifdef` instead?

src/monopole_gr/interp_3d_to_1d.hpp
```diff
@@ -198,29 +212,30 @@ GetMonopoleVarsHelper(const EnergyMomentum &tmunu, const Geometry_t &geom, const
...
-    Real X3 = coords.x1v(k);
-    // TODO(JMM): I kinda don't like this design. Should I redesign
-    // the coordinate transforms in the monopole class to be
-    // self-contained?
-    Geometry::MonopoleCoordTransforms::Cart2Sph(X1, X2, X3, r, th, ph);
-    Geometry::MonopoleCoordTransforms::S2C(r, th, ph, s2c);
-    Geometry::MonopoleCoordTransforms::C2S(X1, X2, X3, r, c2s);
+    const Real X1 = coords.x1v(i);
+    const Real X2 = coords.x1v(j);
+    const Real X3 = coords.x1v(k);
```
brryan, line 226: `x2v` / `x3v`?

brryan (commented):
Nice improvement! I bet time-independent coordinate modifiers will be fine for at least quite a while, maybe forever. A few queries.

cmake/Geometry.cmake
Yurlungur: My thought process here was that there may be some geometries inappropriate for caching. Original the monopole geometries were like this until I added things like time-dependence.

But for example, a geometry that uses the grid through some mechanism that isn't the cache probably shouldn't be cached.

The `MUST_CACHE` variable was originally for `MonopoleCart` until I figured out I didn't need it. We could remove these two lists until we need them. It would make the cmake code a bit simpler.

src/geometry/modified_system.hpp
Yurlungur: `time_dependent` isn't set at compile time... But this gives me an idea. I bet I could make this a template parameter, the same way @lroberts36 does in some other places. Then the compiler should optimize it out.

src/monopole_gr/interp_3d_to_1d.hpp
Yurlungur: yikes good catch.

cmake/Geometry.cmake
Yurlungur: I've removed the `NO_CACHE` and `MUST_CACHE` lists. Let's keep it neat and clean until we need it.

src/geometry/modified_system.hpp
Yurlungur: Can confirm this compiles. Didn't look at the assembly or anything, but I've seen this work elsewhere, so I think we're good.

Yurlungur:
I have added a history reduction into the `tov` package to plot the maximum pressure in the star as a function of time. This will let us see, e.g., quasiperiodic oscillations if we run long enough.

src/phoebus_utils/history.hpp
```diff
@@ -0,0 +1,70 @@
...
+  PARTHENON_REQUIRE(ivar.first >= 0, "Var must exist");
+  PARTHENON_REQUIRE(ivar.second >= ivar.first + idx, "Var must exist");
+
+  Real result = 0.0;
+  Reducer_t reducer(result);
+  parthenon::par_reduce(
+      parthenon::LoopPatternMDRange(), "Phoebus History for " + varname, DevExecSpace(),
+      0, pack.GetDim(5) - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &lresult) {
+        // join is a Kokkos construct
```
Yurlungur, line 59: I made a mistake here. For summations, we need dx^3 which I forgot to include. I will copy what @forrestglines did in Parthenon to make this work automatically.

Yurlungur:
@brryan @jdolence @lroberts36 this is now ready for review. I was able to run a (very) underresolved neutron star on a uniform grid on a single GPU with self gravity, and it's stable. So I call that a win.

In the future it would be good to do some performance optimization, as well as some more strenuous tests and not a uniform grid. I will do that in a future PR. But for the purposes of this PR, I'm pretty happy to declare victory. Some additional changes I needed to implement to get this all working:

1. A bug fix in the sph2cart coordinate transformation to eliminate issues from coordinate singularities
2. Outflow and reflecting boundary conditions (I tested reflecting boundary conditions to implement octant symmetry. It works.)
3. A bug fix in the atmosphere treatment to prevent the atmosphere from diverging at the origin.

src/main.cpp
```diff
@@ -99,3 +81,40 @@ int main(int argc, char *argv[]) {
...
+
+      const std::string bc = pman.pinput->GetOrAddString("phoebus", name, "outflow");
+      if (bc == "reflect") {
+        pman.app_input->boundary_conditions[loc[d - 1][outer]] = reflect[d - 1][outer];
+      } else if (bc == "outflow") {
+        pman.app_input->boundary_conditions[loc[d - 1][outer]] = outflow[d - 1][outer];
+      } // periodic boundaries, which are handled by parthenon, so no need to set anything
+    }
+  }
+}
```
Yurlungur, line 120: @brryan I know you did a similar thing in your PR so we should synchronize.

src/phoebus_boundaries/phoebus_boundaries.cpp
```diff
@@ -31,6 +32,75 @@ using namespace parthenon::package::prelude;
 
 namespace Boundaries {
 
+// Copied out of Parthenon, with slight modification
+enum class BCSide { Inner, Outer };
+enum class BCType { Outflow, Reflect };
+template <CoordinateDirection DIR, BCSide SIDE, BCType TYPE>
+void GenericBC(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
```
Yurlungur, line 39: Jonas's code shamelessly copied out of parthenon

src/fluid/fluid.cpp
```diff
@@ -260,6 +261,33 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+
+  for (int d = 0; d < 3; ++d) {
+    auto ReduceMom = [d](MeshData<Real> *md) {
+      return History::ReduceOneVar<Kokkos::Sum<Real>>(md, fluid_cons::momentum, d);
+    };
+    hst_vars.emplace_back(HistoryOutputVar(
+        HstSum, ReduceMom, "total X" + std::to_string(d + 1) + " momentum"));
+  }
+
+  params.Add(parthenon::hist_param_key, hst_vars);
```
Yurlungur, line 288: Integrated mass, energy, and momentum now all automatically computed in the history file

src/main.cpp
brryan: Yeah I have a `SetPhoebusBoundaries` function in `src/phoebus_boundaries/phoebus_boundaries.cpp`. I think whichever of our PRs gets merged first, this or `add_mocmc`, the other PR can merge in that change, but ultimately I think this function is best housed in `phoebus_boundaries.cpp`

src/phoebus_utils/history.hpp
```diff
@@ -0,0 +1,78 @@
...
+using namespace parthenon::package::prelude;
+
+/*
+ * History Template to build off of. Applies Reducer to MeshData on
+ * variable with varname. If variable is multi-d use idx to offset
+ * from first element. You should capture this object in a lambda to
+ * pass it to the Parthenon history machinery.
+ *
+ * TODO(JMM): This one is very simple. No volume weighting, no metric
+ * information. So it only works for min and max basically.  This can
```
brryan, line 33: doenst it work for volume integrals of densitized conserved quantities?

brryan (approved):
LGTM, thanks for adding history functionality

src/main.cpp
Yurlungur: I can move mine to `phoebus_boundaries.cpp`. We should also make sure our implementations are consistent.

src/phoebus_utils/history.hpp
Yurlungur: oops yes it does. This comment is old.

Yurlungur:
Thanks, @brryan I just pushed addressing your comments. When tests pass, I'll merge.

PR #112, opened 2022-04-29
check for lower-dimensions in interpolation. Resolves #111

### LINKED ISSUE #111: Interpolated cached geometry not working in 1D with particles
When interpolating cached geometry to particle positions for geodesic updates, in 1D I see segfaults because even for a particle in the domain in X2, the geometry caching code will try to access indices at `j=-1`. Some additional checking is needed for lower dimensions, I think

Yurlungur:
## PR Summary

In #111 @brryan pointed out that the cached geometry interpolation segfaults when in 1D because the interpolation operator tries to access invalid memory. This PR resolves the issue by introducing a 1D linear interpolation object. Also, rather than checking for axisymmetry, we now simply check the number of dimensions in the geometry pack. If `GetDim(3) < 2`, then we're axisymmetric If `GetDim(2) < 2`, then we're 1D.

I am still testing, but this is ready for review. @brryan can you try it to see if it resolves your issue?

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/geometry/cached_system.hpp
+++ b/src/geometry/cached_system.hpp
@@ -131,7 +131,7 @@ class Cached {
   }
   KOKKOS_INLINE_FUNCTION
   Real Lapse(int b, Real X0, Real X1, Real X2, Real X3) const {
-    return LCInterp::Do(axisymmetric_, b, X1, X2, X3, pack_, idx_[cent_].alpha);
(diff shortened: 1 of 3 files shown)


Yurlungur:
@brryan this is ready for review

brryan (approved):
Nice fix, thanks for the rapid turnaround, also elegant the way `axisymmetric_` is removed.

PR #114, opened 2022-05-11
Fix TOV

Yurlungur:
## PR Summary

This PR fixes three separate issues I discovered with TOV:
- The names of boundary conditions in the input deck changed between when the TOV was first set up and now. Boundary conditions now correct.
- In the PR where I added time step controls to the monopole solver, I accidentally introduced a bug that set the lapse to 1 always. The issue was the two arrays for the lapse and the lapse at the previous timestep were not being correctly and persistently swapped every cycle. This is now repaired.
- I add floors which can decay with radius, which behave better and cause less perturbations to the star.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/fixup/fixup.hpp
+++ b/src/fixup/fixup.hpp
@@ -14,11 +14,12 @@
 #ifndef FIXUP_FIXUP_HPP_
 #define FIXUP_FIXUP_HPP_
 
+#include <cmath>
+#include <limits>
+
 #include <parthenon/package.hpp>
 #include <utils/error_checking.hpp>
 using namespace parthenon::package::prelude;
-
-#include <limits>
 
 namespace fixup {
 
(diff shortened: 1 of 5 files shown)


src/fixup/fixup.hpp
```diff
@@ -35,47 +36,66 @@ static struct ExpX1RhoSieFloor {
...
+  Floors(RRhoSieFloor, const Real rho0, const Real sie0, const Real rp, const Real sp)
+      : r0_(rho0), s0_(sie0), ralpha_(rp), salpha_(sp), floor_flag_(FloorFlag::RRhoSie) {}
 
   KOKKOS_INLINE_FUNCTION
   void GetFloors(const Real x1, const Real x2, const Real x3, Real &rflr,
                  Real &sflr) const {
-    Real scratch;
     switch (floor_flag_) {
-    case 1:
+    case FloorFlag::ConstantRhoSie:
```
Yurlungur, line 69: Magic numbers now an enum because I can't count to 3.

src/monopole_gr/monopole_gr.cpp
```diff
@@ -416,10 +417,13 @@ TaskStatus SpacetimeToDevice(StateDescriptor *pkg) {
...
   auto alpha = params.Get<Alpha_t>("lapse");
-
-  // Copy old data from alpha into alpha_last
   auto alpha_last = params.Get<Alpha_t>("lapse_last");
+
+  // Swap alpha and alpha last
+  // Params must be updated with new references.
   std::swap(alpha, alpha_last);
+  params.Update("lapse", alpha);
+  params.Update("lapse_last", alpha_last);
```
Yurlungur, line 426: The missing piece needed to make the array swap machinery work

jdolence (approved):

PR #117, opened 2022-06-13
epsmin floor was being set to rhomin floor value in tov pgen

carlnotsagan:
## PR Summary

Fixes tov init bug where `epsmin` floor was being set to `rhomin` value

## PR Checklist

- [ ] Explain what you did.


Yurlungur (approved):
Thanks for the fix!

Issue #126, opened 2022-10-12
Add relevant black hole surfaces to params


Yurlungur:
@AstroBarker pointed out that it would be useful to have the code output critical surfaces like ISCO and outer horizon for black hole spacetimes. I agree, and this would be an easy change. For FMKS, the main black hole spacetime we use, I would suggest adding these to the params in the `Intiialize` function in `geometry/fmks.cpp` [here](https://github.com/lanl/phoebus/blob/main/src/geometry/fmks.cpp#L38). In particular, one would add:
```C++
Real Rh = 1.0 + sqrt(1.0 - a * a);
Real xh = log(Rh);
params.Add("Rh", Rh);
params.Add("xh", xh);
```
and similarly for the radii of the ISCO for co- and counter-rotating disks as described [here](https://en.wikipedia.org/wiki/Innermost_stable_circular_orbit).

@AstroBarker if you're up for it this would be a nice way to get your feet wet with the code. ;)

PR #127, opened 2022-10-13
Blb/issue126 - Add horizon radii and ISCO radii to output

AstroBarker:
Adds horizon radius and prograde/retrograde ISCO radii to `Params/geometry` output -- see Issue  #126.


Yurlungur (approved):
Excellent. Thanks for adding this @AstroBarker :)

PR #134, opened 2022-10-25
Add scattering

brryan:
## PR Summary

This PR adds support for the new (mean) scattering opacity calls in `singularity-opac`. 

It also reorganizes the singularity opacity objects into a single class provided by `phoebus/microphysics/opac_phoebus` to the rest of `phoebus`. 

This PR removes the `scattering_fraction` parameter. 

## PR Checklist

- [x] Explain what you did.


src/pgen/radiation_advection.cpp
```diff
@@ -67,7 +64,7 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
   const Real vx = pin->GetOrAddReal("radiation_advection", "vx", 0.0);
   PARTHENON_REQUIRE(std::fabs(vx) < 1., "Subluminal velocity required!");
   const Real width = pin->GetOrAddReal("radiation_advection", "width", sqrt(2.0));
-  const Real kappa = pin->GetReal("opacity", "gray_kappa") * LENGTH * LENGTH / MASS;
+  const Real kappa = pin->GetReal("s_opacity", "gray_kappa") * LENGTH * LENGTH / MASS;
```
Yurlungur, line 67: What does `s` stand for in `s_opacity`?

src/microphysics/opac_phoebus/opac_phoebus.cpp
```diff
@@ -120,7 +140,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 #endif
   }
 
-  if (do_mean_opacity) {
+  // if (do_mean_opacity) {
```
Yurlungur, line 143: So are we always doing mean opacity now? I guess that isn't unreasonable, since one can just have lots of groups/bins?

src/microphysics/opac_phoebus/opac_phoebus.hpp
```diff
@@ -19,8 +19,112 @@
...
+#include <singularity-opac/neutrinos/s_opac_neutrinos.hpp>
+
 namespace Microphysics {
 
+using RadiationType = singularity::RadiationType;
+
+/// One class to contain all opacity objects and wrap the subset of calls we use in
+/// phoebus for convenience.
+// TODO(BRR) template on opacity/scattering opacity being gray, and if they are constexpr
+// if-away the table lookups for mean opacities
```
Yurlungur, line 36: Should this be in singularity-opac? My feeling is no... and that it belongs in phoebus but I'm curious your thoughts

Yurlungur (approved):
This all seems reasonable to me.

src/pgen/radiation_advection.cpp
Yurlungur: nvm figured this one out lol

src/microphysics/opac_phoebus/opac_phoebus.hpp
brryan: I think I'm actually mistaken in this comment -- the mean opacity table lookups are over rho, T, Ye, *not* frequency, and an opacity that is not only gray but also doesn't depend on `rho`, `T`, `Ye` probably isn't very interesting i.e. we won't use that on large problems where performance is critical. So I'm inclined to just delete this comment and not worry about the case of mean opacity lookups when opacities dont depend on `rho`, `T`, `Ye`

src/pgen/radiation_advection.cpp
brryan: part of me wants to add an `a` to the absorption opacity class for symmetry

Yurlungur: hah either way

src/microphysics/opac_phoebus/opac_phoebus.hpp
```diff
@@ -19,8 +19,110 @@
...
+class Opacities {
+  using Opacity = singularity::neutrinos::Opacity;
+  using MeanOpacity = singularity::neutrinos::MeanOpacity;
+  using SOpacity = singularity::neutrinos::SOpacity;
+  using MeanSOpacity = singularity::neutrinos::MeanSOpacity;
+
+ public:
+  Opacities(const Opacity &opac, const MeanOpacity &m_opac, const SOpacity &s_opac,
+            const MeanSOpacity &m_s_opac)
+      : opac_(opac), m_opac_(m_opac), s_opac_(s_opac), m_s_opac_(m_s_opac) {}
```
brryan, line 44: @Yurlungur this is actually failing on GPUs for me, I'm pretty sure at the first place any opacity functions are called, when calling `Opacities::EnergyDensityFromTemperature` in the torus pgen file, I get a
`return index_ == static_cast<index_t<Ts...>>(-1);` from `variant`. Do you think I need to do anything else to make `Opacities` available on device, like the singularity classes `GetOnDevice {return *this;}` functions?

Yurlungur: Wait are you constructing this `Opacities` object on device? Or constructing it on host and then passing it around? In either case, I can see two possible issues:
1. The default copy constructor is deleted by the compiler if you have const fields. Make your fields non-const (not as args to the constructor, down where you declare them at the bottom of the struct.)
2. Similarly, if you are paccing the `Opacities` struct into kernels because you constructed it on host, the fields can't be references because they need to be captured by value by the lambda.

scripts/python/phoedf.py
```diff
@@ -11,7 +11,11 @@
 # distribute copies to the public, perform publicly and display
 # publicly, and to permit others to do so.
 
-from parthenon_tools import phdf
+import sys
+sys.path.append('/users/brryan/github/phoebus/external/parthenon/scripts/python/packages/parthenon_tools/parthenon_tools/')
```
brryan, line 15: Reminder for me to remove this... I'm having trouble installing `parthenon_tools` on chicoma right now

brryan:
This passes the CI and the rad torus looks good right now so I'm merging

PR #135, opened 2022-10-27
CCSN Initilization

carlnotsagan:
- Reads in 1D stellar profile. 
- Maps to phoebus grid and fills matter variables
- Assumes dr is the same for Monopole solver and input file
compare_KEPLER_and_phoebus_ic

- Has check for `npoints` not equal to input file
- Converts units to scale free assumed by monopole solver.


Yurlungur:
I didn't make it to this yet. Sorry @carlnotsagan I was sick last week and on travel before that. This is on my priority list.

src/ccsn/ccsn.hpp
```diff
@@ -0,0 +1,156 @@
...
+    {
+      inputfile >> val;
+      ccsn_state_raw_h(j, i) = val;
+    }
+  }
+
+  std::cout << "Read 1D profile into state array.\n";
+
+  inputfile.close();
+}
```
Yurlungur, line 124: I think there's probably a way to read the data without looping through the file twice. But that's a problem for future us. This is fine for now.

Yurlungur (commented):
This actually all looks reasonable. I don't see the units issue off the top of my head though I do have some nitpicks. Might need to debug with the code + the input progenitor.

carlnotsagan:
Thanks for these comments. I have been working on chicoma so may need to wait until that is back up. Sounds like there is still a units issue and these are mostly design based. i'll go through them when i can and update once done!

Yurlungur:
I thought I might try to download and poke at it myself to see if I could find the issue. Might ping you for the progenitor file.

PR #141, opened 2022-10-31
Updating torus.cpp to run without radiation

AstroBarker:
After recent updates, the non-radiative torus problem failed when trying to get opacities. I moved all radiation related calls into `if (do_rad) {...}` and am now able to run both `torus.pin` and `torus_rad.pin` without issue.

I also added the `c2p_failure_force_fixup_both = false` flag to the `torus.pin` inputs to accommodate those recent changes.

### DIFF
--- a/src/pgen/torus.cpp
+++ b/src/pgen/torus.cpp
@@ -145,7 +145,6 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
   auto geom = Geometry::GetCoordinateSystem(rc);
 
-
   // set up transformation stuff
   auto gpkg = pmb->packages.Get("geometry");
   bool derefine_poles = gpkg->Param<bool>("derefine_poles");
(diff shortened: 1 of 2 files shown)


src/pgen/torus.cpp
```diff
@@ -277,6 +251,28 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
         // Radiation
         if (do_rad) {
+          StateDescriptor *opac = pmb->packages.Get("opacity").get();
```
brryan, line 254: This won't work on GPU because you're trying to access host side variables (the `Opacities` object in `Params`) inside a compute kernel which can be on GPU. Instead, you have to pull out some `Opacities` object outside of the compute loop and let Kokkos capture that object by value.

AstroBarker: Good to know - thanks!

brryan:
Nice work getting this working, these are the same issues I noticed and fixed in #142, I recommend we merge that branch instead of this because my changes are a bit different

src/pgen/torus.cpp
brryan: :+1: I've been tripped up by this a number of times, and it's something to watch out for especially since we still don't quite have the CI running on gpus.

PR #142, opened 2022-10-31
Fix GRMHD torus

brryan:
I recently broke the GRMHD torus in my #134 PR. This PR fixes that, by providing a default `Opacities()` object in params if radiation is disabled. 

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/fixup/fixup.cpp
+++ b/src/fixup/fixup.cpp
@@ -237,7 +237,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   params.Add("src_failure_strategy", src_failure_strategy);
 
   const bool c2p_failure_force_fixup_both =
-      pin->GetOrAddBoolean("fixup", "c2p_failure_force_fixup_both", true);
+      pin->GetOrAddBoolean("fixup", "c2p_failure_force_fixup_both", false);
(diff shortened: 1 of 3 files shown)


src/fixup/fixup.cpp
```diff
@@ -237,7 +237,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   params.Add("src_failure_strategy", src_failure_strategy);
 
   const bool c2p_failure_force_fixup_both =
-      pin->GetOrAddBoolean("fixup", "c2p_failure_force_fixup_both", true);
+      pin->GetOrAddBoolean("fixup", "c2p_failure_force_fixup_both", false);
```
brryan, line 240: This option is currently broken (race condition) so by default it should be off. There is already a catch elsewhere in the code that crashes if you turn this on.

Yurlungur (approved):
Thanks, @brryan ! Could you also check and see if the changes that @AstroBarker made in #141 make sense? Looks like more if-guards? 

@AstroBarker can you check to see if this fixes things for you? You will need to run `git submodule update --recursive` to force `singularity-opac` to update to include the fixed `Opacity` class.

AstroBarker:
Works for me!

AstroBarker (approved):

PR #143, opened 2022-10-31
Mean opacity constructors expect CGS unit opacities, not NonCGSUnits<>!

brryan:
I noticed that when running outside of `scale_free` mode with radiation, the mean opacities are incorrect. This is because the mean opacity constructors expect an opacity *without* a unit conversion, e.g. `Gray`, whereas in phoebus outside of `scale_free` mode we pass in e.g. `NonCGSUnits<Gray>`.

My solution, to store both CGS and non-CGS host opacities, is a bit yucky. I wonder if this further motivates a `singularity-opac` refactor to deal also with the issue of too many variants brought up by @Yurlungur previously.

## PR Checklist

- [x] Explain what you did.


Yurlungur (approved):
Yeah this is a little gross. I do think we should refactor singularity-opac, but this is fine for now.

PR #145, opened 2022-11-02
Adding mdot to history and moving history to new analysis/ directory

AstroBarker:
I've added the mass accretion rate to Phoebus history output. I also moved `history.hpp` to a new directly, `analysis`. I created `analysis/history_utils.hpp` for calculating the mass flux term and, later, other helper functions as we track more complicated quantities. Below is a quick plot of mdot for a 2D GRMHD torus problem.
image

- [x] Explain what you did.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,6 +187,7 @@ add_subdirectory(src)
 # TODO(JMM): Testing?
 if(PHOEBUS_ENABLE_UNIT_TESTS OR PHOEBUS_ENABLE_INTEGRATION_TESTS OR PHOEBUS_ENABLE_REGRESSION_TESTS)
   message("\nConfiguring tests")
+  enable_testing()
   add_subdirectory(external/catch2 Catch2)
   list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/external/catch2/contrib)
   add_subdirectory(tst)
--- a/src/analysis/history.cpp
+++ b/src/analysis/history.cpp
@@ -69,7 +69,7 @@ Real ReduceMassAccretionRate(MeshData<Real> *md) {
                        (xh - coords.x1v(i)) * m) *
                       dx2 * dx3;
 
-          lresult += - flux;
+          lresult += -flux;
         } else {
           lresult += 0.0;
         }
@@ -78,4 +78,4 @@ Real ReduceMassAccretionRate(MeshData<Real> *md) {
   return result;
 } // mass accretion
 
-} // namespace history
+} // namespace History
--- a/src/analysis/history.cpp
+++ b/src/analysis/history.cpp
@@ -12,6 +12,10 @@
 // publicly, and to permit others to do so.
 
 #include "history.hpp"
+#include "geometry/geometry.hpp"
+#include "geometry/geometry_utils.hpp"
+#include "history_utils.hpp"
+#include "phoebus_utils/relativity_utils.hpp"
 
 namespace History {
 
--- a/src/analysis/history.cpp
+++ b/src/analysis/history.cpp
@@ -1,0 +1,81 @@
+// © 2022. Triad National Security, LLC. All rights reserved.  This
+// program was produced under U.S. Government contract
+// 89233218CNA000001 for Los Alamos National Laboratory (LANL), which
+// is operated by Triad National Security, LLC for the U.S.
+// Department of Energy/National Nuclear Security Administration. All
+// rights in the program are reserved by Triad National Security, LLC,
+// and the U.S. Department of Energy/National Nuclear Security
+// Administration. The Government is granted for itself and others
+// acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works,
+// distribute copies to the public, perform publicly and display
+// publicly, and to permit others to do so.
+
+#include <string>
+#include <vector>
+
+#include <kokkos_abstraction.hpp>
+#include <parthenon/package.hpp>
+#include <utils/error_checking.hpp>
+
+#include "geometry/geometry.hpp"
+#include "geometry/geometry_utils.hpp"
+#include "history.hpp"
+#include "history_utils.hpp"
+#include "phoebus_utils/relativity_utils.hpp"
+#include "phoebus_utils/variables.hpp"
+
+namespace History {
+
+Real ReduceMassAccretionRate(MeshData<Real> *md) {
+  const auto ib = md->GetBoundsI(IndexDomain::interior);
+  const auto jb = md->GetBoundsJ(IndexDomain::interior);
+  const auto kb = md->GetBoundsK(IndexDomain::interior);
+
+  auto pmb = md->GetParentPointer();
+  auto &pars = pmb->packages.Get("geometry")->AllParams();
+  const Real xh = pars.Get<Real>("xh");
+
+  namespace p = fluid_prim;
+  const std::vector<std::string> vars({p::density, p::velocity});
+
+  PackIndexMap imap;
+  auto pack = md->PackVariables(vars, imap);
+
+  const int prho = imap[p::density].first;
+  const int pvel_lo = imap[p::velocity].first;
+  const int pvel_hi = imap[p::velocity].second;
+
+  auto geom = Geometry::GetCoordinateSystem(md);
+
+  Real result = 0.0;
+  parthenon::par_reduce(
+      parthenon::LoopPatternMDRange(), "Phoebus History for Mass Accretion Rate",
+      DevExecSpace(), 0, pack.GetDim(5) - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &lresult) {
+        const auto &coords = pack.GetCoords(b);
+        if (coords.x1f(i) <= xh && xh < coords.x1f(i + 1)) {
+          const Real gdet = geom.DetGamma(CellLocation::Cent, k, j, i);
+
+          const Real dx1 = coords.Dx(X1DIR, k, j, i);
+          const Real dx2 = coords.Dx(X2DIR, k, j, i);
+          const Real dx3 = coords.Dx(X3DIR, k, j, i);
+
+          // interp to make sure we're getting the horizon correct
+          auto m = (CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i + 1) -
+                    CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i - 1)) /
+                   (2.0 * dx1);
+          auto flux = (CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i) +
+                       (xh - coords.x1v(i)) * m) *
+                      dx2 * dx3;
+
+          lresult += - flux;
+        } else {
+          lresult += 0.0;
+        }
+      },
+      result);
+  return result;
+} // mass accretion
+
+} // namespace history
--- a/src/analysis/history.cpp
+++ b/src/analysis/history.cpp
@@ -11,19 +11,7 @@
 // distribute copies to the public, perform publicly and display
 // publicly, and to permit others to do so.
 
-#include <string>
-#include <vector>
-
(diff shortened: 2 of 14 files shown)


src/analysis/history.hpp
```diff
@@ -71,6 +76,59 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
   return result;
 }
 
+template <typename Reducer_t>
```
Yurlungur, line 79: If we're always going to sum, you probably don't need to template on reduction here.

src/analysis/history.hpp
```diff
@@ -71,6 +76,59 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+  auto pack = md->PackVariables(vars, imap);
+
+  const int prho = imap[p::density].first;
+  const int pvel_lo = imap[p::velocity].first;
+  const int pvel_hi = imap[p::velocity].second;
+
+  auto geom = Geometry::GetCoordinateSystem(md);
+
+  Real result = 0.0;
+  Reducer_t reducer(result);
```
Yurlungur, line 102: If we always sum you don't need to instantiate `reducer` you can just leave it as `result`. Sum is the default reducer in kokkos.

src/analysis/history.hpp
```diff
@@ -71,6 +76,59 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+                    CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i - 1)) /
+                   (2.0 * dx1);
+          auto flux = (CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i) +
+                       (xh - coords.x1v(i)) * m) *
+                      dx2 * dx3;
+
+          reducer.join(lresult, -flux);
+        } else {
+          reducer.join(lresult, 0.0);
+        }
```
Yurlungur, line 126: if you are doing sum automatically you can just do `lresult += -flux` and `lresult += 0`

src/analysis/history.hpp
```diff
@@ -71,6 +76,59 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+      },
+      reducer);
```
Yurlungur, line 128: if you are always summing you can just leave this as `result`.

src/analysis/history_utils.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <string>
+#include <vector>
+
+#include "geometry/geometry.hpp"
+#include "geometry/geometry_utils.hpp"
+#include "phoebus_utils/relativity_utils.hpp"
+#include "phoebus_utils/variables.hpp"
+#include <kokkos_abstraction.hpp>
+#include <parthenon/package.hpp>
+#include <utils/error_checking.hpp>
```
Yurlungur, line 26: It doesn't matter much, but I like to space-separate teh parthenon vs phoebus headers. And also, due to lore, I like to order it:
- stdlib headers
- dependencies such as parthenon
- phoebus

so in this case that would look like:
```suggestion
#include <kokkos_abstraction.hpp>
#include <parthenon/package.hpp>
#include <utils/error_checking.hpp>

#include "geometry/geometry.hpp"
#include "geometry/geometry_utils.hpp"
#include "phoebus_utils/relativity_utils.hpp"
#include "phoebus_utils/variables.hpp"
```

src/geometry/geometry.cpp
```diff
@@ -58,6 +51,22 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+    using parthenon::HistoryOutputVar;
+    parthenon::HstVar_list hst_vars = {};
+
+    auto ReduceAccretionRate = [](MeshData<Real> *md) {
+      return ReduceMassAccretionRate<Kokkos::Sum<Real>>(md);
+    };
+
+    hst_vars.emplace_back(HistoryOutputVar(HstSum, ReduceAccretionRate, "mdot"));
+    params.Add(parthenon::hist_param_key, hst_vars);
+  }
```
Yurlungur, line 68: Hmm it occurs to me that something gross about putting this in `geometry::Initialize` is that this machinery is present even if the fluid variables are not. (One could imagine running, e.g., radiation around a black hole, for example.)

One can resolve this (in a kind of hacky way) by also checking the following boolean condition from the input deck
```C++
bool hydro_on = pin->GetBoolean("physics", "hydro");
```

Yurlungur (changes requested):
Thanks for this! Overall this is great! I did notice, however, that this changeset broke unit tests and I think I see why and what the solution is.

Your reduction uses geometry because it pulls in mdot stuff, but some of the geometry machinery is set at configure time. However, that isn't passed into the unit tests, and so the unit tests are complaining that the geometry you've pulled in isn't available.

There are a few solutions:
1. If you make `ReduceMassAccretionRate` not templated on `Reducer_t`, then you can move the implementation into a separate implementation file, e.g., `analysis/history.cpp`. This is safe as far as relocatable device code, because the actual code called inside the `par_reduce` is still "inline." This is the way to go IMO, as it's simple and also will help reduce compile times by compiling the `ReduceMassAccretionRate` function only once.
2. We could pass the configure time information into the unit tests. This would be pretty straightforward, a small cmake tweak or two.
3. We could try and disentangle the history machinery from the unit tests a bit more---it's getting pulled in from the TOV package, because TOV is used as a unit test for the monopole solver. This would be hardest to do but maybe worth it.

I recommend we do option 1. Also a few more comments below:

Yurlungur:
@brryan @bprather also interested in your thoughts on this improvement. :)

AstroBarker:
Thanks for the feedback @Yurlungur. I've de-templated `ReduceMassAccretionRate`as suggested and moved it into `analysis/history.cpp`. I also added a check for if hydro is active to avoid trying to record mdot with no fluid evolution.

src/analysis/history.cpp
```diff
@@ -0,0 +1,69 @@
...
+          const Real dx2 = coords.Dx(X2DIR, k, j, i);
+          const Real dx3 = coords.Dx(X3DIR, k, j, i);
+
+          // interp to make sure we're getting the horizon correct
+          auto m = (CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i + 1) -
+                    CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i - 1)) /
+                   (2.0 * dx1);
+          auto flux = (CalcMassFlux(pack, geom, prho, pvel_lo, pvel_hi, b, k, j, i) +
+                       (xh - coords.x1v(i)) * m) *
+                      dx2 * dx3;
```
brryan, line 58: I don't quite follow this equation -- I agree with it when `xh = coords.x1v(i)`, but otherwise I'm confused. Why not just find the two indices `i` and `i+1` that contain `xh` and then weight their `CalcMassFlux` values by `(xh - coords.x1v(i)) / (coords.x1v(i+1) - coords.x1v(i))` and `(coords.x1v(i+1) - xh) / (coords.x1v(i+1) - coords.x1v(i))` respectively?

src/analysis/history_utils.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  const Real vel[] = {pack(b, pvel_lo, k, j, i), pack(b, pvel_lo + 1, k, j, i),
+                      pack(b, pvel_hi, k, j, i)};
+
+  Real gcov4[4][4];
+  geom.SpacetimeMetric(CellLocation::Cent, k, j, i, gcov4);
+  const Real W = phoebus::GetLorentzFactor(vel, gcov4);
+  const Real ucon = vel[0] - shift[0] * W / lapse;
+
+  return gdet * pack(b, prho, k, j, i) * ucon;
```
brryan, line 51: Do we actually want `lapse * gdet * pack(b, prho, k, j, i) * ucon;`? I think we want the integral of the mass flux over the spherical shell at `i` which I understand to be `\int \sqrt{-g} \rho u^1 dx2 dx3` (the `dx2 dx3` show up later of course). Whereas here `gdet = DetGamma = \sqrt{-g} / \alpha`.

brryan (changes requested):
Nice work @AstroBarker! Just some small queries.

src/analysis/history_utils.hpp
AstroBarker: Good catch, I missed that `gdet` was `det gamma`. Fixing this now.

src/analysis/history.cpp
Yurlungur: I think that's equivalent... This version is an interpolation of that right?

src/analysis/history.hpp
```diff
@@ -21,6 +21,12 @@
 #include <parthenon/package.hpp>
 #include <utils/error_checking.hpp>
 
+#include "geometry/geometry.hpp"
+#include "geometry/geometry_utils.hpp"
+#include "history_utils.hpp"
+#include "phoebus_utils/relativity_utils.hpp"
+#include "phoebus_utils/variables.hpp"
```
Yurlungur, line 28: Move as many of these include statements as possible out of the header file and into `analysis/history.hpp`. I suspect you can at least move `#include "geometry/geometry.hpp"` and `i#include "phoebus_utils/relativity_utils.hpp"` into `analysis/history.hpp`

Yurlungur (changes requested):
Thanks for the fixes! The unit tests are still failing because of the include statements below. We should move as many of them as we can into `history/analysis.cpp`.

In general it's good practice to minimize the number of includes in header files, as those includes get recompiled everywhere, and they also introduce additional inter-dependency between source files, which can easily result in circular dependencies.

src/analysis/history.cpp
AstroBarker: So I used the linear interpolation for `y(xh) = y0 + (xh - x0) m` for slope `m`, which is equivalent to the weighted average form of linear interpolation. A slight difference perhaps, for the slope I used two neighboring cells `i+1` and `i-1`.

src/analysis/history.hpp
AstroBarker: Done, thanks!

CMakeLists.txt
```diff
@@ -187,6 +187,7 @@ add_subdirectory(src)
 # TODO(JMM): Testing?
 if(PHOEBUS_ENABLE_UNIT_TESTS OR PHOEBUS_ENABLE_INTEGRATION_TESTS OR PHOEBUS_ENABLE_REGRESSION_TESTS)
   message("\nConfiguring tests")
+  enable_testing()
```
Yurlungur, line 190: Needed to get `ctest` to properly register unit tests for some reason.

tst/CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ if (PHOEBUS_ENABLE_MPI)
     MPI_PARALLEL)
 endif()
 
+if(PHOEBUS_ENABLE_UNIT_TESTS)
+  message("Building unit tests.")
+  add_subdirectory(unit)
+endif()
```
Yurlungur, line 32: don't know if this change matters. But it doesn't hurt.

tst/unit/geometry/CMakeLists.txt
```diff
@@ -39,6 +41,10 @@ add_library(geometry_unit_tests OBJECT
 
 )
 
+# Super annoying but it had to be done.
+set(MAX_NUMBER_CONSERVED_VARS 9 CACHE STRING "Maximum number of conserved variables")
+configure_file(../../../src/compile_constants.hpp.in generated/compile_constants.hpp @ONLY)
+
```
Yurlungur, line 47: These were the relevant changes for making the relevant file get autogenerated.

Yurlungur (approved):
Tests were still failing, so I gave up and generated the relevant file and hooked it into tests. Also fixed a bizarre error where ctest wasn't properly slurping up tests.

Thanks for all your work on this @AstroBarker and for putting up with our nitpicking. As soon as @brryan approves let's merge this in. :)

src/analysis/history.cpp
brryan: Yeah youre right sorry, I just wasnt thinking clearly.

Yurlungur:
@brryan if you're happy with the PR now can you approve?

brryan (approved):

PR #146, opened 2022-11-07
Get torus working with restarts

brryan:
## PR Summary

We weren't able to properly restart torus simulations with radiation due to the higher-dimensional radiation quantities. This PR fixes that, based on the changes to parthenon in https://github.com/parthenon-hpc-lab/parthenon/pull/781. In this PR this mostly required changing `phoedf` to accommodate the new output format. 

This also adds a maximum optical depth per timestep as an input parameter. The default is `1e5`. In some cases very very large optical depths can cause numerical failures in the source terms. 

## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Format your changes by calling `scripts/bash/format.sh`.
- [x] Explain what you did.
- [x] Merge https://github.com/parthenon-hpc-lab/parthenon/pull/781 into `parthenon`
- [x] Merge https://github.com/lanl/singularity-opac/pull/43 into `singularity-opac`

PR #147, opened 2022-11-09
Fix radiation stress energy tensor in radiation moment geometric sources

brryan:
## PR Summary

I was using an incorrect expression for the radiation stress energy tensor when evaluating the radiation moment geometric source terms (Basically I was treating the spatial projection of the four-vector `Halpha` as equivalent to `Halpha`). 

I also switched the expression for the geometric source terms in the radiation moments to match what we already use in the fluid (the two expressions should be equivalent but it would be weird to report different forms of the equations in e.g. a methods paper. I also anyway had a bug of an extra factor of `alpha` on one term in my old version).

I also switched some manual `for` loops in the fluid geometric sources to our `SPACETIMELOOP*` macros.

## PR Checklist

- [x] Explain what you did.
- [ ] Check that the no-B-fields radiative Fishbone solution still looks good
- [ ] Check that the B-fields GRRMHD torus still looks good

### DIFF
--- a/scripts/python/plot_torus_3d.py
+++ b/scripts/python/plot_torus_3d.py
@@ -82,6 +82,7 @@ def plot_frame(ifname, fname, savefig, geomfile=None, rlim=40, coords='cartesian
   if ndim == 2:
     xblock = rblock*np.sin(thblock)
     yblock = rblock*np.cos(thblock)
+    zblock = None
   elif ndim == 3:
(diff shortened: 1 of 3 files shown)


brryan:
There's too much thrashing with submodules right now for me to easily/reliably test this branch, so I put these changes into #146, tested them, and they seem to work fine.

PR #151, opened 2022-12-01
Add magnetic flux Phi to history

AstroBarker:
Here I add the magnetic flux Phi to the output. Below shows the unnormalized and normalized magnetic flux. The scales seem okay but it's hard for me to compare directly to the plots in the EHT code comparison paper because they do 3D and the dynamics are a bit different. Thoughts?
image
image

- [x] Explain what you did.

### DIFF
--- a/src/analysis/history.cpp
+++ b/src/analysis/history.cpp
@@ -186,4 +186,57 @@ Real ReduceJetMomentumFlux(MeshData<Real> *md) {
 
 } // ReduceJetMomentumFlux
 
+Real ReduceMagneticFluxPhi(MeshData<Real> *md) {
+  const auto ib = md->GetBoundsI(IndexDomain::interior);
+  const auto jb = md->GetBoundsJ(IndexDomain::interior);
+  const auto kb = md->GetBoundsK(IndexDomain::interior);
+
+  auto pmb = md->GetParentPointer();
+  auto &pars = pmb->packages.Get("geometry")->AllParams();
+  const Real xh = pars.Get<Real>("xh");
+
+  namespace p = fluid_prim;
+  const std::vector<std::string> vars({p::bfield, p::velocity});
+
+  PackIndexMap imap;
+  auto pack = md->PackVariables(vars, imap);
+
+  const int pvel_lo = imap[p::velocity].first;
+  const int pvel_hi = imap[p::velocity].second;
+  const int pb_lo = imap[p::bfield].first;
+  const int pb_hi = imap[p::bfield].second;
+
+  auto geom = Geometry::GetCoordinateSystem(md);
+
+  Real result = 0.0;
+  parthenon::par_reduce(
+      parthenon::LoopPatternMDRange(), "Phoebus History for Jet Momentum Flux",
+      DevExecSpace(), 0, pack.GetDim(5) - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &lresult) {
+        const auto &coords = pack.GetCoords(b);
+        if (coords.x1f(i) <= xh && xh < coords.x1f(i + 1)) {
+          const Real dx1 = coords.Dx(X1DIR, k, j, i);
+          const Real dx2 = coords.Dx(X2DIR, k, j, i);
+          const Real dx3 = coords.Dx(X3DIR, k, j, i);
+
+          // interp to make sure we're getting the horizon correct
+          auto m = (CalcMagneticFluxPhi(pack, geom, pvel_lo, pvel_hi, pb_lo, pb_hi, b, k,
+                                        j, i + 1) -
+                    CalcMagneticFluxPhi(pack, geom, pvel_lo, pvel_hi, pb_lo, pb_hi, b, k,
+                                        j, i - 1)) /
+                   (2.0 * dx1);
+          auto flux = (CalcMagneticFluxPhi(pack, geom, pvel_lo, pvel_hi, pb_lo, pb_hi, b,
+                                           k, j, i) +
+                       (xh - coords.x1v(i)) * m) *
+                      dx2 * dx3;
+
+          lresult += flux;
+        } else {
+          lresult += 0.0;
+        }
+      },
+      result);
+  return 0.5 * result;
+} // Phi
+
 } // namespace History
--- a/src/analysis/history.cpp
+++ b/src/analysis/history.cpp
@@ -212,16 +212,17 @@ Real ReduceMagneticFluxPhi(MeshData<Real> *md) {
       KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &lresult) {
         const auto &coords = pack.GetCoords(b);
         if (coords.x1f(i) <= xh && xh < coords.x1f(i + 1)) {
-          const Real dx1 = coords.Dx(X1DIR, k, j, i+1);
-          const Real dx2 = coords.Dx(X2DIR, k, j, i+1);
-          const Real dx3 = coords.Dx(X3DIR, k, j, i+1);
+          const Real dx1 = coords.Dx(X1DIR, k, j, i + 1);
+          const Real dx2 = coords.Dx(X2DIR, k, j, i + 1);
+          const Real dx3 = coords.Dx(X3DIR, k, j, i + 1);
 
           // interp to make sure we're getting the horizon correct
-          auto m = (CalcMagneticFluxPhi(pack, geom, cb_lo, b, k, j, i + 1 +1) -
-                    CalcMagneticFluxPhi(pack, geom, cb_lo, b, k, j, i - 1 +1)) /
+          auto m = (CalcMagneticFluxPhi(pack, geom, cb_lo, b, k, j, i + 1 + 1) -
+                    CalcMagneticFluxPhi(pack, geom, cb_lo, b, k, j, i - 1 + 1)) /
                    (2.0 * dx1);
-          auto flux = (CalcMagneticFluxPhi(pack, geom, cb_lo, b, k, j, i+1) +
-                       (xh - coords.x1v(i+1)) * m) * dx2 * dx3;
+          auto flux = (CalcMagneticFluxPhi(pack, geom, cb_lo, b, k, j, i + 1) +
+                       (xh - coords.x1v(i + 1)) * m) *
+                      dx2 * dx3;
 
           lresult += flux;
         } else {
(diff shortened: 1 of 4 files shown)


src/analysis/history_utils.hpp
```diff
@@ -154,6 +154,38 @@ KOKKOS_INLINE_FUNCTION Real CalcEMMomentumFlux(Pack &pack, Geometry &geom,
...
+  for (int m = 0; m < 3; m++)
+    for (int n = 0; n < 3; n++) {
+      Bdotv += gam[m][n] * Bp[m] * robust::ratio(vcon[n], W);
+    }
+  const Real ucon0 = W / lapse;
+  const Real ucon1 = vcon[0] - shift[0] * W / lapse;
+  const Real b0 = Bdotv * W / lapse;
+  const Real bcon1 = robust::ratio((Bp[0] + lapse * b0 * ucon1), W);
+
+  return std::abs(bcon1 * ucon0 - b0 * ucon1) * lapse * gdet;
```
Yurlungur, line 186: I don't think this is right. In equation 14 of the EHT paper, `*F^{rt}` isn't the MHD flux in the r,t direction... It's the `r,t` components of the hodge star of the Maxwell tensor. And in fact, we have

*F^{it} = B^i

so you should just integrate sqrt{-g}*B^r here, rather than computing primitive B.

src/analysis/history.cpp
```diff
@@ -186,4 +186,57 @@ Real ReduceJetMomentumFlux(MeshData<Real> *md) {
 
 } // ReduceJetMomentumFlux
 
+Real ReduceMagneticFluxPhi(MeshData<Real> *md) {
```
Yurlungur, line 189: This pattern is almost identical to the other reduce patterns. I wonder if this can be made generic as something like
```C++
template<typename Function>
Real ReduceHorizonQuantity(MeshData<Real> *md, Function &f) {
  // ...

  Real result = 0.0;
  parthenon::par_reduce(
      parthenon::LoopPatternMDRange(), "Phoebus History for Jet Momentum Flux",
      DevExecSpace(), 0, pack.GetDim(5) - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &lresult) {
        const auto &coords = pack.GetCoords(b);
        if (coords.x1f(i) <= xh && xh < coords.x1f(i + 1)) {
          const Real dx1 = coords.Dx(X1DIR, k, j, i);
          const Real dx2 = coords.Dx(X2DIR, k, j, i);
          const Real dx3 = coords.Dx(X3DIR, k, j, i);

          // interp to make sure we're getting the horizon correct
          auto m = (f(pack, geom, pvel_lo, pvel_hi, pb_lo, pb_hi, b, k, j, i + 1) -
                    f(pack, geom, pvel_lo, pvel_hi, pb_lo, pb_hi, b, k, j, i - 1)) /
                   (2.0 * dx1);
          auto flux = (f(pack, geom, pvel_lo, pvel_hi, pb_lo, pb_hi, b, k, j, i) +
                       (xh - coords.x1v(i)) * m) *
                      dx2 * dx3;

          lresult += flux;
        } else {
          lresult += 0.0;
        }
      },
      result);
  return 0.5 * result;
}
```
something like that? Then we could combine the reduction functions for flux with mdot etc. Anything that's a horizon quantity could have a single reduction function? What do you think?

Yurlungur (changes requested):
Something seems off here. Here's what comes out of nubhlight in 2d:
image

src/analysis/history_utils.hpp
brryan: I think `*F^{it} = B^i` is true for the Harm `B^i`, but since in our code we differ by a factor of alpha don't we need `*F^{it} = B^i / alpha` in this expression?

src/analysis/history.cpp
bprather: So, I tried something very like this: take a look at KHARMA's reductions/reductions.hpp [here](https://github.com/AFD-Illinois/kharma/blob/dev/kharma/reductions/reductions.hpp), which has my best shot at it.

The major difference is, I don't force the reduction body to be a given function type, instead letting it be an arbitrary code block, at the cost of doing some truly unspeakable things with macros.  The flexibility is nice, but it was only really crucial back in the old pre-packing days when I absolutely needed the lambda capture to work.  I'm no longer sure it's a better way to go than the single f(pass, the, whole, world, as, arguments) approach, which I admittedly use in plenty of other places in KHARMA, but there may be some extra use/flexibility I'm not thinking of, and certainly adding new reductions has a lot less boilerplate in the macro version.

Honestly, though, if a function template solution really works, I'd love to help with & adopt something like that in KHARMA.

AstroBarker: I had been thinking about this too and wasn't sure the cleanest way to do it. I'll play with this in a separate PR

src/analysis/history_utils.hpp
Yurlungur: oh yes good call. Still either way---it's not b^i

AstroBarker:
I've changed the `Phi` integrand to be `Bp[0] * gdet` (so `Bp[0] / lapse * (lapse * detgamma)`). This is how the unnormalized `Phi` looks now.. Still quite different from nublight?
image

src/analysis/history.cpp
```diff
@@ -186,4 +186,57 @@ Real ReduceJetMomentumFlux(MeshData<Real> *md) {
...
+                       (xh - coords.x1v(i)) * m) *
+                      dx2 * dx3;
+
+          lresult += flux;
+        } else {
+          lresult += 0.0;
+        }
+      },
+      result);
+  return 0.5 * result;
```
Yurlungur, line 239: Why is there a `0.5` here?

src/analysis/history_utils.hpp
```diff
@@ -154,6 +154,24 @@ KOKKOS_INLINE_FUNCTION Real CalcEMMomentumFlux(Pack &pack, Geometry &geom,
   return -T_EM_11 * lapse * gdet;
 } // momentum flux
 
+template <typename Pack, typename Geometry>
+KOKKOS_INLINE_FUNCTION Real CalcMagneticFluxPhi(Pack &pack, Geometry &geom,
+                                                const int pvel_lo, const int pvel_hi,
```
Yurlungur, line 159: Don't need `pvel_lo` and `pvel_hi` anymore.

src/analysis/history.cpp
AstroBarker: Eq 14 for `\Phi_{BH}` in the EHT paper has it https://arxiv.org/pdf/1904.04923.pdf

src/analysis/history_utils.hpp
```diff
@@ -154,6 +154,24 @@ KOKKOS_INLINE_FUNCTION Real CalcEMMomentumFlux(Pack &pack, Geometry &geom,
...
+  Real gam[3][3];
+  geom.Metric(CellLocation::Cent, k, j, i, gam);
+  Real gdet = geom.DetGamma(CellLocation::Cent, k, j, i);
+  Real lapse = geom.Lapse(CellLocation::Cent, k, j, i);
+
+  const Real Bp[] = {pack(b, pb_lo, k, j, i), pack(b, pb_lo + 1, k, j, i),
+                     pack(b, pb_hi, k, j, i)};
+
+  // \int B^r sqrt(-gdet) := B^r / alpha * (alpha * detgamma)
+  return std::abs(Bp[0]) * gdet;
```
Yurlungur, line 172: The primitive `B` field in Phoebus is densitized---it's multiplied by a factor of `detgam`. So you don't actually need to multiply here by `gdet`. I think that's what's wrong. 

I *think* that that implies this whole function could just be replaced with `return std::abs(pack(b, pb_lo, k, j, i));`

src/analysis/history_utils.hpp
```diff
@@ -154,6 +154,24 @@ KOKKOS_INLINE_FUNCTION Real CalcEMMomentumFlux(Pack &pack, Geometry &geom,
...
+                                                const int pb_lo, const int pb_hi,
+                                                const int b, const int k, const int j,
+                                                const int i) {
+  Real gam[3][3];
+  geom.Metric(CellLocation::Cent, k, j, i, gam);
+  Real gdet = geom.DetGamma(CellLocation::Cent, k, j, i);
+  Real lapse = geom.Lapse(CellLocation::Cent, k, j, i);
+
+  const Real Bp[] = {pack(b, pb_lo, k, j, i), pack(b, pb_lo + 1, k, j, i),
+                     pack(b, pb_hi, k, j, i)};
```
Yurlungur, line 169: Can clean this up to not define unused variables.

Yurlungur (changes requested):
I think I see what's wrong.

src/analysis/history.cpp
Yurlungur: Ah ok. We might want to put that in the integrand then, rather than in the final integral. That might make things cleaner in the long run. But no big deal either way.

src/analysis/history_utils.hpp
AstroBarker: Ahhh I see. So is the primitive `detgam B / alpha`, regarding the conversations above?

Yurlungur: Ah... I think I've confused myself. Looking again at the phoebus doc:
- Primitive variable is just $B^i$
- Conserved variable is $\sqrt{\gamma} B^i$

and this is the difference @brryan was referring to but he got it backwards. In HARM the conserved variable is 

I'm now looking at `prim2con` and the phoebus doc. The HARM conserved variable is $\sqrt{g} B^i$, but as far as I can tell the primitive variables are identical. I think this is what @brryan was referring to before.

So if you use the `fluid_cons` instead of `fluid_prim` then all you need to do is multiply by the lapse to get $\sqrt{g}B^i$ and then you're done. I.e.,
```C++
// assumes that pack is packing conserved B not primitive B
// conserved B is \sqrt{\gamma} B^k
// we want \sqrt{g} B^k = \sqrt{gamma} alpha B^k
return std::abs(pack(b, pb_lo, k, j, i))*lapse;
```

Sorry for the confusion! I think me and @brryan spouted half truths that made sense in our head without sitting down and writing it out

Yurlungur: Also... nice latex works now in github. When did that happen?

AstroBarker: Ah perfect! Will implement these tomorrow

Yurlungur: Edited to take all the blame instead of @brryan sorry Ben. ;)

AstroBarker:
After #153 we have the following Phi, Phi/sqrt(mdot), and mdot.
image
image
image

Yurlungur:
That looks much better!

Yurlungur (approved):
I think this is ready to go. Thanks for chasing this down, @AstroBarker !

PR #153, opened 2023-01-11
Fix: update restart logic in main.cpp

AstroBarker:
There was a logic issue / missing `!` in `main.cpp` that prevented the post initialization modifications from being applied on the first run. Should be the source of the weird Phi behavior seen in  #151 as well.

- [ x] Format your changes by calling `scripts/bash/format.sh`.
- [x ] Explain what you did.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -60,7 +60,7 @@ int main(int argc, char *argv[]) {
   pman.ParthenonInitPackagesAndMesh();
 
   // call post-initialization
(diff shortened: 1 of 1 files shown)


Yurlungur:
Thanks for the catch here! We should make sure @brryan has eyes on this too to make sure this matches his intent.

brryan (approved):
Nice catch @AstroBarker, thanks!

PR #155, opened 2023-02-08
Mg/homologouscollapse

mari2895:
## PR Checklist

- [ ] Explain what you did.


Tests/Reader/converted_ADM_stellartable.dat
```diff
@@ -0,0 +1,2001 @@
+# r mass_density temp Ye specific_internal_energy velocity pressure adm_density adm_momentum S_adm Srr_adm
```
Yurlungur, line 1: This file and the other table should probably only be committed once, in the top-level `data` directory`

Tests/Reader/Makefile
```diff
@@ -0,0 +1,6 @@
+ALL: compile run
+compile: reader.cpp main.cpp
+	g++ -o compile reader.cpp main.cpp -I.
```
Yurlungur, line 3: This version of the makefile doesn't use patterns. You could automatically generate rules to create object files like this:
```
%.o: %.cpp reader.hpp
	g++ -o $@ $^ -I.
```
The `%.o: %.cpp` creates a rule to generate an object file for each `cpp` file that depends on `header.hpp`. In our case it resolves to
- `reader.o: reader.cpp reader.hpp`
- `main.o: main.cpp reader.hpp`

Then the script line in the tab to generate the target uses automatic variables. The `$@` means the target. So in this case the `.o` file. The `$^` means all dependencies. So the `.cpp` file and the `reader.hpp` file.

Then you will need a linker line such as

```
reader: reader.o main.o
	g++ -o $@ $^
```

Tests/Reader/Makefile
```diff
@@ -0,0 +1,6 @@
...
+
+run: reader.cpp main.cpp
+	./compile "converted_ADM_stellartable.dat"
```
Yurlungur, line 6: The targets here are wrong. You want to only run after the reader has been compiled so it should look like this:
```
run: reader
	./reader "converted_ADM_stellartable.dat"
```

Tests/Reader/Makefile
```diff
@@ -0,0 +1,6 @@
+ALL: compile run
```
Yurlungur, line 1: With the other changes I suggest in the file, this becomes
```
ALL: reader run
```

Tests/Reader/Makefile
Yurlungur: It's also good practice to add two special targets to your make file.

The first is `clean` which removes build artifacts:
```
clean:
	$(RM) reader reader.o main.o
```
where $(RM) resolves to `rm -f` by default.

and 
```
.PHONY: all run clean
```
which tells gnu make that the `all`, `run`, and `clean` targets don't actually generate files they're just steps in the build process.

Yurlungur: One more thing, you can assign variables just like a shell script. So you can for example specify 
```
CC:=g++
```
to specify the compiler should be `g++` and then use `$(CC)` everywhere you would use `g++`. This is often good practice to make your makefile more flexible.

Tests/Reader/reader.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <iostream>
+#include <vector>
+#include <string>
+#include <sstream>
+#include <cmath>
+using namespace std;
+
+namespace COLUMNS {
+  constexpr int rad = 0, rho = 1, temp=2, ye = 3, eps = 4, vel = 5, press = 6, rho_adm = 7, P_adm =8, S_adm = 9, Srr_adm = 10;
+ }
```
Yurlungur, line 11: You might want to move this to `reader.hpp` so it's available everywhere you might want it.

Tests/Reader/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+cmake_minimum_required(VERSION 3.10)
+project (reader)
+include_directories(/home/mari/Work/Phoebus/phoebus/Tests/Reader/)
+add_executable (output /home/mari/Work/Phoebus/phoebus/Tests/Reader/reader.cpp /home/mari/Work/Phoebus/phoebus/Tests/Reader/main.cpp /home/mari/Work/Phoebus/phoebus/Tests/Reader/reader.hpp)
```
Yurlungur, line 4: This look like absolute paths. As a general rule, use relative paths when you can. Otherwise this file will only work on your computer and won't be portable.

Yurlungur (changes requested):
Nice work.

When you're done playing with things here, we might want to re-organize a bit. For example when the reader is connected to `Phoebus` it probably should live in `phoebus_utils` or something similar. No need to wory about it for now though.

src/pgen/homologous.cpp~
```diff
@@ -0,0 +1,132 @@
+// © 2021. Triad National Security, LLC. All rights reserved.  This
```
Yurlungur, line 1: This is a backup file generated by emacs. You should remove it from the version control with `git rm`. We should also add `*.cpp~` to the `.gitignore` file at the top of the repo.

inputs/homologous.pin
```diff
@@ -0,0 +1,146 @@
...
+# time step will be <= dtfac*(time rate of change of lapse)
+dtfac = 0.9
+# warn_on_dt tries to estimate if first-order operator split is sufficiently accurate.
+# Compares (dalpha/dt) computed analytically with numerical difference
+# between two subcycles. Warns if difference is larger than dtwarn_eps
+warn_on_dt = false
+dtwarn_eps = 1e-5
+
+<homologous>
+table="/home/mari/Work/Phoebus/phoebus/Tests/Reader/converted_ADM_homologouscollapse.dat"
```
Yurlungur, line 146: Try to make this a relative path

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,137 @@
...
+// acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works,
+// distribute copies to the public, perform publicly and display
+// publicly, and to permit others to do so.
+
+#include <cmath>
+#include "geometry/geometry.hpp"
+#include "monopole_gr/monopole_gr.hpp"
+#include "pgen/pgen.hpp"
+#include "/Work/Phoebus/phoebus/Tests/GRsolver/reader.hpp"
```
Yurlungur, line 18: you don't have enough cmake for the linker to link correctly yet because the cmake is all in yoru `Tests` directory. You need to hook into the rest of teh phoebus build system

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,137 @@
...
+  const int iye = imap[fluid_prim::ye].second;
+  const int iprs = imap[fluid_prim::pressure].first;
+  const int itmp = imap[fluid_prim::temperature].first;
+  const int igm1 = imap[fluid_prim::gamma1].first;
+
+
+  
+  const Real Cv = pin->GetOrAddReal("homologous", "Cv", 1.0);
+  const Real gamma = pin->GetOrAddReal("homologous", "gamma", 1.3334);
+  const bool spherical = pin->GetOrAddReal("homologous", "spherical_coords", true);
```
Yurlungur, line 53: If you're pulling htese parameters from the input deck, you need to add them to the input deck. Easier is probably pulling htem from the eos block of the input deck
```C++
  const Real Cv = pin->GetOrAddReal("eos", "Cv", 1.0);
  const Real gamma = pin->GetOrAddReal("eos", "gamma", 1.3334);
```

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,137 @@
...
+  const string table = pin ->GetOrAddReal("homologous","table");
+  auto &coords = pmb->coords;
+  auto pmesh = pmb->pmy_mesh;
+  int ndim = pmesh->ndim;
+
+  const Real v_inner = (4. / 3.) * M_PI * std::pow(rinner, 3.);
```
Yurlungur, line 59: where is `rinner` first defined?

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,137 @@
...
+  MonopoleGR::LinearSolveForAlpha(monopole_pkg);
+  MonopoleGR::SpacetimeToDevice(monopole_pkg);
+  MonopoleGR::DumpToTxt("phoebus-metric-setup.dat",monopole_pkg);
+
+  // Primitive Variables
+  auto MoveVectorToDevice(std::vector<double> &vec){
+    Kokkos::View<Real*, Kokkos::HostSpace, Kokkos::MemoryUnmanaged> hm (vec.data(),vec.size());
+    return;
+    Kokkos::create_mirror_view_and_copy(Kokkos::DefaultExecutionSpace::memory_space(),hm);
+  }
```
Yurlungur, line 98: You can't have nested function definitions in C++ in the same way you can in python. This needs to be a fucntion above the `pgen` declaration. OR you can define it inline as a lambda:
```C++
  auto MoveVectorToDevice = [](std::vector<double> &vec) {
    Kokkos::View<Real*, Kokkos::HostSpace, Kokkos::MemoryUnmanaged> hm (vec.data(),vec.size());
    return;
    Kokkos::create_mirror_view_and_copy(Kokkos::DefaultExecutionSpace::memory_space(),hm);
  }
```

Yurlungur (commented):
Some additional comments

src/pgen/homologous.cpp
mari2895: Fixed

src/pgen/homologous.cpp
mari2895: commented out

src/pgen/homologous.cpp
mari2895: defined inline as lambda

src/pgen/homologous.cpp
mari2895: Okay so I changed this to Tests/Reader/reader.hpp. I added CMakelists.txt in the Tests directory which has add_subdirectory of Reader in it. I also added corresponding link in phoebus's CMakelist.txt.

inputs/homologous.pin
mari2895: changed to "../Tests/Reader/converted_ADM_homologouscollapse.dat"

src/pgen/homologous.cpp
Yurlungur: :+1: does it compile now?

mari2895: No it says Monopole GR is only enabled with TOV.

CMakeLists.txt
```diff
@@ -184,6 +184,9 @@ include(cmake/Fluid.cmake)
 message("\nConfiguring src")
 add_subdirectory(src)
 
+# Testing / Reader
+add_subdirectory(Tests)
+
```
Yurlungur, line 189: With the re-org to move most of this code into the `src` directory, is this needed anymore?

Data/converted_ADM_stellartable.dat
```diff
@@ -0,0 +1,2001 @@
+# r mass_density temp Ye specific_internal_energy velocity pressure adm_density adm_momentum S_adm Srr_adm
```
Yurlungur, line 1: As we discussed, we should remove this file, and instead provide instructions for attaining a progenitor file and then generating the ADM version.

Tests/Reader/CMakeLists.txt
Yurlungur: I assume that with the reader integrated in phoebus, this test can go away.

Alternatively, we can add the test to the phoebus unit tests. I can explain how to do that.

Tests/Reader/Makefile
Yurlungur, line 1: This makefile can go away after code reorg

inputs/progenitor.pin
```diff
@@ -0,0 +1,166 @@
...
+dtwarn_eps = 1e-5
+
+<units>
+scale_free = false
+fluid_mass_g = 1.3140924486113762e+37
+geom_mass_g = 1.3140924486113762e+37
+
+<progenitor>
+enabled=true
+tablepath=converted_ADM_stellartable.dat
```
Yurlungur, line 166: add a newline at the end of this line

src/fluid/prim2con.hpp
```diff
@@ -24,7 +24,7 @@ namespace prim2con {
 KOKKOS_INLINE_FUNCTION
 void signal_speeds(const Real &rho, const Real &u, const Real &p, const Real &bsq,
                    const Real v[], const Real &vsq, const Real &gam1, const Real &alpha,
-                   const Real beta[], const Real gcon[3][3], Real *sig) {
+                   const Real beta[], const Real gcon[3][3], Real *sig) {
```
Yurlungur, line 27: Looks like you added a trailing space here, which we don't need. No big deal either way though

src/monopole_gr/interp_3d_to_1d.hpp
```diff
@@ -147,8 +147,8 @@ TaskStatus InterpolateMatterTo1D(Data *rc) {
         // paranoid. Need to make sure we don't need miss any cells in
         // the 1d grid.
         // No correctness issue from making this too big,
-        int i1dleft = radius1d.index(r - dr) - 1;
-        int i1dright = radius1d.index(r + dr) + 1;
+	int i1dleft = std::max(0,radius1d.index(r - dr) - 1);
+        int i1dright = std::min(radius1d.index(r + dr) + 1, npoints-1);
```
Yurlungur, line 151: :+1:

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,204 @@
...
+  PARTHENON_REQUIRE(is_monopole_cart || is_monopole_sph, "Monopole GR Required");
+  auto &rc = pmb->meshblock_data.Get();
+  auto geom=Geometry::GetCoordinateSystem(rc.get());
+    
+  PackIndexMap imap;
+  auto v = rc->PackVariables(
+      {fluid_prim::density, fluid_prim::velocity, fluid_prim::energy, fluid_prim::bfield,
+       fluid_prim::ye, fluid_prim::pressure, fluid_prim::temperature, fluid_prim::gamma1},
+      imap);
+  std::cout<< "PackVariablesCompleted"<<std::endl;
```
Yurlungur, line 44: We should delete this debug message

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "monopole_gr/monopole_gr.hpp"
+#include "microphysics/eos_phoebus/eos_phoebus.hpp"
+#include "pgen/pgen.hpp"
+#include "phoebus_utils/ascii_reader.hpp"
+#include "progenitor/progenitor.hpp"
+
+// Homologously collapsing star.
+namespace homologous {
+  void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
+    std::cout << "Entered into ProbGen" <<std::endl;
```
Yurlungur, line 30: We should delete this debug message

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,204 @@
...
+  const int iye = imap[fluid_prim::ye].second;
+  const int iprs = imap[fluid_prim::pressure].first;
+  const int itmp = imap[fluid_prim::temperature].first;
+  const int igm1 = imap[fluid_prim::gamma1].first;
+
+
+  // DO WE NEED THOSE?
+  const Real Cv = pin->GetOrAddReal("eos", "Cv", 1.0);
+  const Real gamma = pin->GetOrAddReal("eos", "gamma", 1.3334);
+  //const bool spherical = pin->GetOrAddReal("homologous", "spherical_coords", true);
```
Yurlungur, line 60: We do not. This whole block can be deleted.

src/pgen/homologous.cpp
```diff
@@ -0,0 +1,204 @@
...
+  
+  auto &coords = pmb->coords;
+  auto pmesh = pmb->pmy_mesh;
+  int ndim = pmesh->ndim;
+
+  //const Real v_inner = (4. / 3.) * M_PI * std::pow(rinner, 3.);
+  //const Real uinner = Eexp / v_inner;
```
Yurlungur, line 67: We can delete these.

src/pgen/progenitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+    matter_h(MonopoleGR::Matter::trcS,i)=S_adm.interpToReal(rad.x(i));
+    matter_h(MonopoleGR::Matter::Srr,i)=Srr_adm.interpToReal(rad.x(i));
+  }
+  Kokkos::deep_copy(matter,matter_h);
+  MonopoleGR::IntegrateHypersurface(monopole_pkg.get());
+  MonopoleGR::LinearSolveForAlpha(monopole_pkg.get());
+  MonopoleGR::SpacetimeToDevice(monopole_pkg.get());
+  MonopoleGR::DumpToTxt("phoebus-metric-setup.dat",monopole_pkg.get());
+  monopole_initialized = true;
+  } // if statement
```
Yurlungur, line 109:
```suggestion
  } // if (!monopole_initialized)
```

src/pgen/progenitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+	const Real u = phoebus::energy_from_rho_P(eos, mass_density_dev.interpToReal(r),pressure_dev.interpToReal(r),emin,emax,lambda[0]);
+	const Real sie = u/mass_density_dev.interpToReal(r);
+        const Real T = eos.TemperatureFromDensityInternalEnergy(mass_density_dev.interpToReal(r), sie, lambda);
+
+	Real vel_vec_in[3]={0};
+	Real vel_vec_out[3]={0};
+	vel_vec_in[0] = velocity_dev.interpToReal(r);
+	if (is_monopole_cart){
+	  for ( int i=0; i<3; ++i){
+	    for (int j=0; j<3; ++j){
```
Yurlungur, line 148: I'm not sure if this is more or less clear, but Phoebus provides a macro for 3D loops. You could replace these two loops with
```C++
SPACELOOP2(i, j) {
```
and then you would need one fewer close `}` below as well.

src/pgen/progenitor.cpp
```diff
@@ -0,0 +1,180 @@
...
+	      vel_vec_out[i]+=s2c[i][j]*vel_vec_in[j];
+	    }
+	  }
+	}
+	else {
+	  vel_vec_out[0] = vel_vec_in[0];
+	}
+	
+        v(irho, k, j, i) = mass_density_dev.interpToReal(r);
+        for (int d = 0 ; d < 3; ++d){
```
Yurlungur, line 158:
```suggestion
        SPACELOOP(d) {
```

src/phoebus_driver.cpp
```diff
@@ -684,9 +688,9 @@ parthenon::Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
   if (enable_tov && !enable_monopole) {
     PARTHENON_THROW("MonopoleGR required for TOV initialization");
   }
-  if (enable_monopole && !enable_tov) {
-    PARTHENON_THROW("Currently monopole GR only enabled with TOV");
-  }
+  //if (enable_monopole && !enable_tov) {
+  //  PARTHENON_THROW("Currently monopole GR only enabled with TOV");
+  //}
```
Yurlungur, line 693:
```suggestion
const auto enable_progenitor = progenitor_pkg->Param<bool>("enabled");
  if (enable_monopole && !(enable_tov || enable_progenitor)) {
    PARTHENON_THROW("Currently monopole GR only enabled with TOV or Progenitor packages");
  }
```

src/phoebus_driver.cpp
```diff
@@ -704,6 +708,7 @@ parthenon::Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
     }
   }
   return packages;
+
```
Yurlungur, line 711: unnecessary newline

src/phoebus_utils/ascii_reader.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <cmath>
+#include <utils/error_checking.hpp>
+#include "ascii_reader.hpp"
+
+using namespace std;
+namespace AsciiReader {
+
+void readtable (string filename, std::vector<double> mydata[11]){
+
+  //std::vector<double> mydata[11];
```
Yurlungur, line 15: no need to keep this commented out line

src/phoebus_utils/ascii_reader.cpp
```diff
@@ -0,0 +1,51 @@
...
+ double calculatemass(std::vector<double> mydata[11]){
+ double mass=0;
+ for (int i=0; i<=mydata[0].size()-1; i++){
+   //mass=mass+4*3.14*(rho[i+1]*pow(rad[i+1],2)+rho[i]*pow(rad[i],2))/2*(rad[i+1]-rad[i]);
+   std::vector<double> &rho = mydata[COLUMNS::rho];
+   std::vector<double> &rad = mydata[COLUMNS::rad];
+   mass=mass+4*3.14*(rho[i+1]*pow(rad[i+1],2)+rho[i]*pow(rad[i],2))/2*(rad[i+1]-rad[i]);
+ }
+ //std::cout << mass << std::endl;
+
```
Yurlungur, line 46: No need to keep these empty lines

Yurlungur (changes requested):
Thanks for all your hard work on this, @mari2895 . There's a lot here, and it was a lot of work to get to this point. Some minor comments/nitpicks below, but also some big picture comments/questions:
1. Is your Python code in the repo? We should put it into `script/python`
2. Before final merge, we should run the auto-formatter. There's a script in `script/bash/format.sh`. See if you can run that script. You may be missing some dependencies. If so, let's talk about it over slack.
3. to delete a file from git it's not enough to just remove the file, you need to type `git rm filename`.
4. We should test your reader with a unit test in the catch framework.

CMakeLists.txt
mari2895: I don't think so

scripts/python/homologous/in_solution.py
```diff
@@ -0,0 +1,137 @@
...
+import time
+import matplotlib.pyplot as pl
+import sys
+import glob
+from scipy.integrate import odeint
+
+
+G=6.69e-8
+c=3.e10
+msun=2.e33
```
Yurlungur, line 14: Units/constants package? Also you should probably use higher precision for these numbers.

src/fluid/prim2con.hpp
```diff
@@ -29,7 +29,7 @@ void signal_speeds(const Real &rho, const Real &u, const Real &p, const Real &bs
   const Real vasq = bsq / (rho_rel + bsq);
   Real cssq = gam1 * p / rho_rel;
   PARTHENON_DEBUG_REQUIRE(rho > robust::SMALL(), "rho is unacceptably small!");
-  PARTHENON_DEBUG_REQUIRE(u > robust::SMALL(), "u is unacceptably small!");
+  //PARTHENON_DEBUG_REQUIRE(u > robust::SMALL(), "u is unacceptably small!");
```
Yurlungur, line 32: Can you try uncommenting this out and running in debug mode? We may not need to comment it out anymore

Yurlungur (changes requested):
Thanks for all these changes, @mari2895 a few more things and then I think this is good to go:
1. It looks like tests are failing and the code is not compiling. Did you add the changes we made to prevent the issue with `compile_constants` and push that?
2. After the final changes are done, can you try running `script/bash/format.sh` and committing the files it changes?

Yurlungur:
@lroberts36 @brryan @jdolence @carlnotsagan  this MR adds everything needed to read a progenitor. Heads up I plan to merge soon.

scripts/python/grsolver/grsolver.py
```diff
@@ -0,0 +1,319 @@
...
+
+                for i in range(len(r)):
+                        #upper metric
+                        g00=-1./alpha[i]**2.   
+                        g0r=beta[i]/alpha[i]**2.                
+                        gamma2=1./(1-self.v_int(r[i])**2./c**2.)
+                        rho_adm[i]=alpha[i]**2.*((self.rho_int(r[i])+self.p_int(r[i])/c**2.)*gamma2-self.p_int(r[i])/c**2.*g00)
+                        P_adm[i]=alpha[i]*beta[i]*((self.rho_int(r[i])+self.p_int(r[i])/c**2.)*gamma2+self.p_int(r[i])/c**2.*g00)+alpha[i]*((self.rho_int(r[i])+self.p_int(r[i])/c**2.)*gamma2*self.v_int(r[i])+self.p_int(r[i])/c**2.*g0r)   ## r-component  (upper index)
+                        S_adm[i]=(alpha[i]**2.*gamma2-1.)*self.rho_int(r[i])*c**2.+(alpha[i]**2.*gamma2+2.)*self.p_int(r[i])
+                        Srr_adm[i]=self.rho_int(r[i])*self.v_int(r[i])**2.*gamma2*a[i]**4.+self.p_int(r[i])*a[i]**4.*(1./a[i]**2-beta[i]**2./alpha[i]**2.+self.v_int(r[i])**2./c**2.*gamma2)
```
AstroBarker, line 224: This code is getting a bit difficult to read.. it might be good practice for us to adopt a code formatting strategy for python like we do with `clang-format`. I think `autopep8` is popular, even if I am not the hugest pep8 fan.

scripts/python/grsolver/seos.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+
+def FindTemperature(rho,p,lmbda,tmin,tmax,eos):
+        def f(x):
+            return eos.PressureFromDensityTemperature(rho,x,lmbda)-p
+        sol=bisect(f,tmin,tmax)
+        #T=sol.x[0]
+        T=sol
+        #print('T=',T, 'rho=',rho,'p=',p)
```
AstroBarker, line 18: probably safe to remove commented out code here?

scripts/python/grsolver/convert.py
```diff
@@ -0,0 +1,41 @@
...
+    newdata['adm_momentum']=data['adm_momentum']*G**3./c**7.*M0**2.
+    newdata['S_adm']=data['S_adm']*G**3./c**8.*M0**2.
+    newdata['Srr_adm']=data['Srr_adm']*G**3./c**8.*M0**2.
+    ascii.write(newdata,'converted_'+filename,overwrite=True,format=formattype)
+    return
+
+
+filename='ADM_stellartable.dat'
+#filename='ADM_homologouscollapse.dat'
+#filename='ADM_tov.dat'
```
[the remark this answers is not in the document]
AstroBarker: With this we could add a `-h` help flag for new folks trying to run the scripts.

scripts/python/grsolver/grsolver.py
```diff
@@ -0,0 +1,319 @@
...
+                rho_adm=interp1d(self.R,rho_adm0,fill_value='extrapolate')(r)
+                P_adm=interp1d(self.R,P_adm0,fill_value='extrapolate')(r)
+                S_adm=interp1d(self.R,S_adm0,fill_value='extrapolate')(r)
+                Srr_adm=interp1d(self.R,Srr_adm0,fill_value='extrapolate')(r)
+                a=interp1d(self.R,a0,fill_value='extrapolate')(r)
+                K=interp1d(self.R,K0,fill_value='extrapolate')(r)
+                alpha=interp1d(self.R,alpha0,fill_value='extrapolate')(r)
+                #np.save('a',a)
+                #np.save('K',K)
+                #np.save('alpha',alpha)
```
AstroBarker, line 270: Removable lines?

AstroBarker (commented):
Looks good to me. My suggestions are small. Since we are using `astropy` already, we could use their `astropy.constants` resource for physical constants (comes with all the full accepted precision, error proof). I also think we should adopt some python formatting strategy, such as `autopep8`.

Yurlungur:
> I also think we should adopt some python formatting strategy, such as autopep8.

Parthenon uses `black` so we could adopt that. I'll make an issue--I think that should be a separate PR.

src/pgen/progenitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+#include "geometry/geometry.hpp"
+#include "monopole_gr/monopole_gr.hpp"
+#include "microphysics/eos_phoebus/eos_phoebus.hpp"
+#include "pgen/pgen.hpp"
+#include "phoebus_utils/ascii_reader.hpp"
+#include "progenitor/progenitordata.hpp"
+
+// Homologously collapsing star.
+namespace homologous {
```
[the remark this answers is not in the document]
Yurlungur: So in most places, namespaces should be capitalized but because of insane macro magic, `pgen` is special. The namespace needs to be exactly the name you want for the problem when requesting it in the input deck.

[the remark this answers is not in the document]
Yurlungur: i.e., this one is fine. :)

scripts/python/grsolver/grsolver.py
Yurlungur: I added an issue #163 . That should be a separate MR at a later date.

scripts/python/grsolver/grsolver.py
```diff
@@ -0,0 +1,320 @@
...
+from scipy.interpolate import interp1d
+from scipy.integrate import odeint
+from scipy.integrate import solve_ivp 
+from scipy import optimize
+from numpy import diff
+from seos import CalculateInternalEnergy
+from argparse import ArgumentParser
+
+G=const.G.cgs.value
+c=const.c.cgs.value
```
Yurlungur, line 23: :+1:

src/radiation/cooling_function.cpp
```diff
@@ -71,6 +73,113 @@ TaskStatus CoolingFunctionCalculateFourForce(MeshBlockData<Real> *rc, const doub
         v(Gye, k, j, i) = 0.;
       });
 
+  // Light Bulb with Liebendorfer model
+  auto &coords = pmb->coords;
+  const bool do_liebendorfer = rad->Param<bool>("do_liebendorfer");
+  const bool do_lightbulb = rad->Param<bool>("do_lightbulb");
+  const Real lum = rad->Param<Real>("lum");
+  if (do_lightbulb) {
```
Yurlungur, line 81: For future reference, these changes should have been made in a separate branch to keep changesets self-contained. It's fine for now though.

Yurlungur (approved):
This looks good to me now. I'll format it and merge pending you figure otu what's going on with the outer boundary.

Then we can merge again once lightbulb is fully working after we pull in the mass fractions data we need.

src/radiation/radiation.cpp
```diff
@@ -112,6 +112,16 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
   params.Add("absorption", absorption);
 
+  if (method == "cooling_function") {
+    const bool do_liebendorfer =
+        pin->GetOrAddBoolean("radiation", "do_liebendorfer", false);
+    const bool do_lightbulb = pin->GetOrAddBoolean("radiation", "do_lightbulb", false);
+    const Real lum = pin->GetOrAddReal("radiation", "lum", 4.0e52);
+    params.Add("radiation", "do_liebendorfer");
+    params.Add("radiation", "do_lightbulb");
+    params.Add("radiation", "lum");
```
Yurlungur, line 122: These lines are causing the tests to fail. Syntax is:
```suggestion
    params.Add("do_liebendorfer", do_liebendorfer);
    params.Add("do_lightbulb", do_lightbulb);
    params.Add("lum", lum);
```
`params` is already owned by the `radiation` package.

Yurlungur (approved):

PR #156, opened 2023-02-24
Fix FMKS

brryan:
## PR Summary

FMKS, unlike MKS, is broken on the torus problem -- in 2D the disk blows up after about 200M. 

This PR aims to fix that.

~~At the moment it looks like the issue was a bug in the singularity fix code (only being applied to theta = 0 but not theta = pi, using `abs` instead of `fabs`)~~
This PR fixes unrelated bugs in the polar axis singularity fixup code and in using function arguments for `GetNorm_`

It turns out the problem is when reflecting the X2 component of the magnetic field about the polar axis in FMKS. Why this plagues phoebus but not harm is currently unknown. Outflow BCs for X2 work fine, and this is probably a moot point once we switch to face-centered fields.

We add a `polar` BC option for X2 but all this does is enable a special fix flux option; otherwise it looks just like reflection in `GenericBC`. 

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/geometry/fmks.cpp
+++ b/src/geometry/fmks.cpp
@@ -60,6 +60,16 @@ void Initialize<FMKSMeshBlock>(ParameterInput *pin, StateDescriptor *geometry) {
   params.Add("h", h);
   params.Add("xt", xt);
   params.Add("alpha", alpha);
+  // Require x0 to be <= x1 of the innermost ghost cell face to avoid issues with mapping theta 
+  // outside of [0, pi]
+  // This calculation assumes no mesh refinement, but works for any refinement because ghost cell
+  // layers of refined blocks will always have smaller extents
+  const Real x1min =  pin->GetReal("parthenon/mesh", "x1min");
+  const Real x1max =  pin->GetReal("parthenon/mesh", "x1max");
+  const int nx1 = pin->GetInteger("parthenon/mesh", "nx1");
+  const Real dx1 = (x1max - x1min) / nx1;
+  const int nghost = pin->GetInteger("parthenon/mesh", "nghost");
+  x0 = std::min<Real>(x0, std::max<Real>(0., x1min - nghost * dx1));
   params.Add("x0", x0);
   params.Add("smooth", smooth);
   params.Add("Rh", Rh);
--- a/src/geometry/fmks.cpp
+++ b/src/geometry/fmks.cpp
@@ -60,12 +60,12 @@ void Initialize<FMKSMeshBlock>(ParameterInput *pin, StateDescriptor *geometry) {
   params.Add("h", h);
   params.Add("xt", xt);
   params.Add("alpha", alpha);
-  // Require x0 to be <= x1 of the innermost ghost cell face to avoid issues with mapping theta 
-  // outside of [0, pi]
-  // This calculation assumes no mesh refinement, but works for any refinement because ghost cell
-  // layers of refined blocks will always have smaller extents
-  const Real x1min =  pin->GetReal("parthenon/mesh", "x1min");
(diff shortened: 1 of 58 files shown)


Yurlungur (commented):
Oops this is almost certainly a bug I introduced. Sorry about that and thanks for the fix. Is FMKS still broken after these fixes?

brryan:
> Oops this is almost certainly a bug I introduced. Sorry about that and thanks for the fix. Is FMKS still broken after these fixes?

Yes this does not fix the FMKS issue unfortunately. I'll keep looking

src/phoebus_boundaries/phoebus_boundaries.cpp
```diff
@@ -136,6 +143,72 @@ void OutflowInnerX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
...
+  const auto idx_pb = imap.GetFlatIdx(fluid_prim::bfield, false);
+
+  pmb->par_for_bndry(
+      "PolarInnerX2Prim", nb, domain, coarse,
+      KOKKOS_LAMBDA(const int &l, const int &k, const int &j, const int &i) {
+        const int jref = -j + 2 * j0 - 1;
+        if (l == idx_pvel(1)) {
+          q(l, k, j, i) = -q(l, k, jref, i);
+        } else if (l == idx_pb(1)) {
+          q(l, k, j, i) = -q(l, k, jref, i);
```
brryan, line 172: We could use this function (which right now behaves just like the generic BC for prim vars) to implement @Yurlungur's suggestion of only flipping the velocity component. Doing that leads to a stable torus.

src/geometry/fmks.cpp
```diff
@@ -60,6 +60,16 @@ void Initialize<FMKSMeshBlock>(ParameterInput *pin, StateDescriptor *geometry) {
...
+  // Require x0 to be <= x1 of the innermost ghost cell face to avoid issues with mapping
+  // theta outside of [0, pi] This calculation assumes no mesh refinement, but works for
+  // any refinement because ghost cell layers of refined blocks will always have smaller
+  // extents
+  const Real x1min = pin->GetReal("parthenon/mesh", "x1min");
+  const Real x1max = pin->GetReal("parthenon/mesh", "x1max");
+  const int nx1 = pin->GetInteger("parthenon/mesh", "nx1");
+  const Real dx1 = (x1max - x1min) / nx1;
+  const int nghost = pin->GetInteger("parthenon/mesh", "nghost");
+  x0 = std::min<Real>(x0, std::max<Real>(0., x1min - nghost * dx1));
```
Yurlungur, line 72: nice fix

src/geometry/mckinney_gammie_ryan.hpp
```diff
@@ -147,26 +149,40 @@ class McKinneyGammieRyan {
...
       thJ_(X2, y, thJ);
-      th = thG + std::exp(smooth_ * (x0_ - X1)) * (thJ - thG);
+      const Real fac = x0_ - X1;
+      th = thG + std::exp(smooth_ * fac) * (thJ - thG);
     } else {
       y = 0;
       th = thG;
     }
     // coordinate singularity fix at the poles. Avoid theta = 0.
-    if (std::abs(th) < robust::EPS()) th = robust::sgn(th) * robust::EPS();
```
Yurlungur, line 156: oops thanks for the fix

src/phoebus_boundaries/phoebus_boundaries.cpp
Yurlungur: :+1:

Yurlungur (approved):
Thanks for digging into this

PR #158, opened 2023-03-02
Fix restriction bug on initialization

Yurlungur:
## PR Summary

Upon updating parthenon, @brryan found that Phoebus was crashing when restriction was called. The issue turned out to be a subtle interaction between `Metadata` and the custom prolongation/restriction operations. Essentially primitive  and/or conserved variables need to be set for prolongation/restriction when they are used for boundary conditions. However, they were being set with an "empty" prolongation/restriction operator and the code was crashing when the stored functor was a `nullptr`.

This was happening because because `Metadata::FillGhost` was being set after the constructor for the metadata object was called, but the default prolongation/restriction ops are currently only set when `Metadata` is constructed. 

I fix this issue here by creating `Metadata` objects once and not changing their flags after construction. There should also be some changes in `parthenon` to guard against this in the future. But this fixes Phoebus for now.

I also tweak the `eos_phoebus.cpp` file while I was debugging. Essentially I just moved the vector of valid names outside `initialize`.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/fluid/fluid.cpp
+++ b/src/fluid/fluid.cpp
@@ -131,40 +131,43 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   Real sigma_cutoff = pin->GetOrAddReal("fluid", "sigma_cutoff", 1.0);
   params.Add("sigma_cutoff", sigma_cutoff);
 
-  Metadata m;
   std::vector<int> three_vec(1, 3);
+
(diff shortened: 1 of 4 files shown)


brryan (approved):
Nice fix @Yurlungur, thanks for cleaning up my eos leftovers, I'll merge this when CI completes.

For posterity, worth noting that this issue did NOT appear in the *2D* torus, only the *3D* torus.

PR #159, opened 2023-03-14
Get torus working with nuclear EoS

AstroBarker:
This PR adds necessary functionality to get the torus problem working with StellarCollapse EoS.

## PR Summary
I add the following functionalities:
- `adiabats.hpp` which includes necessary functions to compute adiabats from the table given S, Ye.
  - A unit test for this has been added.
- In the torus pgen, I pull in the various quantities needed (adiabatic S, Ye, table bounds). The calculations for `rho` and `u` are replaced by `GetStateFromEnthalpy` which returns `rho` and `u` for either the ideal gas or stellar collapse EoS.

Misc change: bug found in `root_find` where `status` was not passed through correctly and it complained when the root find failed even when status was passed in.

## TODO:
- @brryan When you have time, can you verify that this change does not break things for your applications? (Note: You may need to update `singularity-eos` to its latest and checkout `jhp/entropy`.
- I would like to unify the adiabat framework to work for the ideal gas as well, but I need to reason around this `kappa`.
- @brryan Do you know of anything immediately obvious that I would need to change for this to work with radiation? I haven't modified any of those parts of the pgen yet.
- This requires the `jhp/entropy` branch of `singularity-eos`. Once that is through I need to update the used version of singularity to latest commit. @jonahm-LANL I tried committing `singularity-eos` to latest but I think it did not stick. What is the way to do this?

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/inputs/torus.pin
+++ b/inputs/torus.pin
@@ -153,4 +153,4 @@ entropy = 4
 <units>
 scale_free = false
 geom_mass_msun = 2.8
-fluid_density_g = 1.e11
+fluid_density_cgs = 1.e11
--- a/inputs/torus.pin
+++ b/inputs/torus.pin
@@ -82,11 +82,10 @@ method = derivative_order_1
 max_level = 3
 
 <eos>
-# type = IdealGas
-type = StellarCollapse
-filter_bmod = false
-filename = ../../Data/sfho.h5
-use_sp5 = false
+type = IdealGas
+#type = StellarCollapse
+#filename = ../../Data/sfho.h5
+#use_sp5 = false
 Gamma = 1.66666666666666666666666667
 Cv = 1.0
 
@@ -143,7 +142,8 @@ n_inside_horizon = 5
 magnetized = true
 entropy = 4
 
-<units>
-scale_free = false
-geom_mass_msun = 2.8
-fluid_mass_g = 1.e26
+# For tabulated EoS
+#<units>
+#scale_free = false
+#geom_mass_msun = 2.8
+#fluid_density_g = 1.e11
--- a/inputs/torus.pin
+++ b/inputs/torus.pin
@@ -94,6 +94,7 @@ Gamma = 1.66666666666666666666666667
 Cv = 1.0
 rho_min = 1e-8
 sie_min = 1e-8
+nsamps_adiabat = 180
 
 <physics>
 hydro = true
--- a/inputs/torus.pin
+++ b/inputs/torus.pin
@@ -25,6 +25,9 @@ variables = p.density,       &
             c.energy,        &
             pressure,      	 &
             g.c.coord,       &
+            g.f1.coord,       &
+            g.f2.coord,       &
+            g.f3.coord,       &
             g.n.coord,       &
             g.c.detgam,      &
             g.c.bcon,        &
@@ -82,12 +85,15 @@ method = derivative_order_1
 max_level = 3
 
 <eos>
-type = IdealGas
-#type = StellarCollapse
-#filename = ../../Data/sfho.h5
-#use_sp5 = false
+#type = IdealGas
+type = StellarCollapse
+filename = ../../Data/sfho.h5
+use_sp5 = false
+filter_bmod = false
 Gamma = 1.66666666666666666666666667
 Cv = 1.0
+rho_min = 1e-8
+sie_min = 1e-8
 
 <physics>
 hydro = true
@@ -103,7 +109,7 @@ recon = weno5
 c2p_max_iter = 100
 c2p_tol = 1.e-8
 c2p_method = robust
-Ye = true
+Ye = false
 mhd = true
 
 <geometry>
@@ -143,7 +149,7 @@ magnetized = true
 entropy = 4
 
 # For tabulated EoS
-#<units>
-#scale_free = false
-#geom_mass_msun = 2.8
-#fluid_density_g = 1.e11
+<units>
+scale_free = false
+geom_mass_msun = 2.8
+fluid_density_g = 1.e11
--- a/inputs/torus.pin
+++ b/inputs/torus.pin
@@ -82,7 +82,11 @@ method = derivative_order_1
 max_level = 3
 
 <eos>
-type = IdealGas
+# type = IdealGas
+type = StellarCollapse
+filter_bmod = false
+filename = ../../Data/sfho.h5
+use_sp5 = false
 Gamma = 1.66666666666666666666666667
 Cv = 1.0
 
@@ -100,7 +104,7 @@ recon = weno5
 c2p_max_iter = 100
 c2p_tol = 1.e-8
 c2p_method = robust
-Ye = false
+Ye = true
 mhd = true
 
 <geometry>
@@ -137,3 +141,9 @@ rin = 6.0
 rmax = 12.0
 n_inside_horizon = 5
 magnetized = true
+entropy = 4
+
+<units>
+scale_free = false
+geom_mass_msun = 2.8
+fluid_mass_g = 1.e26
--- a/inputs/torus.pin
+++ b/inputs/torus.pin
@@ -85,11 +85,10 @@ method = derivative_order_1
 max_level = 3
 
 <eos>
-#type = IdealGas
-type = StellarCollapse
-filename = ../../Data/sfho.h5
-use_sp5 = false
-filter_bmod = false
+type = IdealGas
+#type = StellarCollapse
+#filename = ../../Data/sfho.h5
+#use_sp5 = false
 Gamma = 1.66666666666666666666666667
 Cv = 1.0
 rho_min = 1e-8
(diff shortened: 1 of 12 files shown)


AstroBarker:
Also the logic to make it work with IdealGas is kind of hacky.. I would like to fix this in the future by unifying the adiabat framework to work for an arbitrary EoS (without crash) but I need to better understand the `kappa` business first, and also this PR is already quite long..

Yurlungur:
@AstroBarker looks like the singularity-eos submodule this is pinned to doesn't contain the entropy... Maybe something funny with `git submodule`s?

Yurlungur:
> @AstroBarker looks like the singularity-eos submodule this is pinned to doesn't contain the entropy... Maybe something funny with `git submodule`s?

Oh I just saw your comment above. Not sure what's going on here. I'll chat with you on slack

brryan:
@AstroBarker This should work fine with the existing radiative torus setup. To do that, we take in a GRMHD solution with known density/temperature everywhere, and then in each zone treat the GRMHD pressure as the total pressure and then solve the equation P_tot = P_gas(T) + P_rad(T) for T, where T will in general be less than the initial GRMHD solution's temperature. As long as the EOS used by your PR and that gas/rad partitioning step are the same (they should be of course) this should be fine. 

I will try to test for my use case soon, and at least review/approve ~tomorrow.

Yurlungur:
@brryan we're currently facing a GPU bug with singularity-eos. Trying to track it down.

Yurlungur:
@AstroBarker where is this as far as being able to merge it in?

AstroBarker:
> @AstroBarker where is this as far as being able to merge it in?

I've got some local changes that I need to finish testing and push. The changes try to use the adiabat framework for the ideal gas after converting `kappa` -> `entropy`. Letting the root find go to much lower rho, T like we talked about helps but it's still a bit fickle to get `rho_rmax` the same, but is closer than before.

Yurlungur:
> > @AstroBarker where is this as far as being able to merge it in?
> 
> I've got some local changes that I need to finish testing and push. The changes try to use the adiabat framework for the ideal gas after converting `kappa` -> `entropy`. Letting the root find go to much lower rho, T like we talked about helps but it's still a bit fickle to get `rho_rmax` the same, but is closer than before.

How far off are they from each other now?

AstroBarker:
> > > @AstroBarker where is this as far as being able to merge it in?
> > 
> > 
> > I've got some local changes that I need to finish testing and push. The changes try to use the adiabat framework for the ideal gas after converting `kappa` -> `entropy`. Letting the root find go to much lower rho, T like we talked about helps but it's still a bit fickle to get `rho_rmax` the same, but is closer than before.
> 
> How far off are they from each other now?

Only a couple percent, so long as things are set correctly. It took some playing with e.g., how close to 0 I let the density/temp go on the adiabats and the number of samples used.

Yurlungur:
> > > > @AstroBarker where is this as far as being able to merge it in?
> > > 
> > > 
> > > I've got some local changes that I need to finish testing and push. The changes try to use the adiabat framework for the ideal gas after converting `kappa` -> `entropy`. Letting the root find go to much lower rho, T like we talked about helps but it's still a bit fickle to get `rho_rmax` the same, but is closer than before.
> > 
> > 
> > How far off are they from each other now?
> 
> Only a couple percent, so long as things are set correctly. It took some playing with e.g., how close to 0 I let the density/temp go on the adiabats and the number of samples used.

Hmm. That's probably fine... but since the torus is a standard test problem, it's probably useful to be able to retain the old original ideal gas setup as well.

How gross would it be to include both that machinery and a runtime flag that reverts to the old analytic formula? Not set on the eos type but something like `torus/use_legacy_init` or something?

AstroBarker:
Have elected to keep the switches on EoS in the torus init. Made a few changes to root find tolerances. I think this is good to go. There may need to be some additional work when it is time for radiation, but that can be done at that time.

Yurlungur:
Looks like tests are failing:
https://github.com/lanl/phoebus/actions/runs/4622700920/jobs/8175695524?pr=159#step:5:344

inputs/torus.pin
```diff
@@ -100,7 +103,7 @@ recon = weno5
 c2p_max_iter = 100
 c2p_tol = 1.e-8
 c2p_method = robust
-Ye = false
+Ye = true
```
Yurlungur, line 106: maybe still set default to false so as to not mess up the ideal gas run. I guess Ye doesn't do anything in that case.

src/pgen/torus.cpp
```diff
@@ -183,15 +238,44 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
   RNGPool rng_pool(seed);
 
+  int nsamps = 180; // for adiabats
+  if (eos_type != "StellarCollapse") {
+    nsamps = 1;
+  }
```
Yurlungur, line 244: Maybe make `nsamps` an input parameter that defaults to 180?

src/pgen/torus.cpp
```diff
@@ -530,6 +619,39 @@ Real log_enthalpy(const Real r, const Real th, const Real a, const Real rin, con
   return lnh;
 }
 
+/**
+ * cheeky way to make this pgen work for ideal and nuclear EoS
+ * There's probably a better way to do this
+ **/
```
Yurlungur, line 625: Nah this is fine.

src/phoebus_utils/adiabats.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <kokkos_abstraction.hpp>
+#include <utils/error_checking.hpp>
+
+#include <spiner/databox.hpp>
+
+#include "microphysics/eos_phoebus/eos_phoebus.hpp"
+#include "phoebus_utils/root_find.hpp"
+
+// a namespace?
+using Microphysics::EOS::EOS;
```
Yurlungur, line 29: As a general rule, it's not great to have a `using` statement outside a namespace in a header file. Maybe hide all of this behind an `Adiabats` namespace or something similar?

src/phoebus_utils/adiabats.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+template <typename EOS>
+inline void GetRhoBounds(const EOS &eos, const Real rho_min, const Real rho_max,
+                         const Real T_min, const Real T_max, const Real Ye, const Real S0,
+                         Real &lrho_min_new, Real &lrho_max_new) {
+
+  // adjust bounds slightly to avoid edges
+  const Real ADJUST = 0.01; // 0.01
```
Yurlungur, line 37:
```suggestion
  constexpr Real ADJUST = 0.01;
```

src/phoebus_utils/root_find.hpp
```diff
@@ -112,7 +112,7 @@ struct RootFind {
                                            RootFindStatus *status = nullptr) {
     Real ya, yb;
     refine_bracket(func, guess, a, b, ya, yb);
-    if (!check_bracket(a, b, ya, yb)) {
+    if (!check_bracket(a, b, ya, yb, status)) {
```
Yurlungur, line 115: :+1:

src/phoebus_utils/unit_conversions.cpp
```diff
@@ -35,12 +35,17 @@ UnitConversions::UnitConversions(ParameterInput *pin) {
     int geom_mass_g_exists = pin->DoesParameterExist("units", "geom_mass_g");
     int geom_mass_msun_exists = pin->DoesParameterExist("units", "geom_mass_msun");
     int geom_length_cm_exists = pin->DoesParameterExist("units", "geom_length_cm");
+    int fluid_density_g_exists = pin->DoesParameterExist("units", "fluid_density_g");
```
Yurlungur, line 38: `_g` here implies grams, rather than g/cm^3. maybe `_cgs`?

Yurlungur (approved):
Thanks @AstroBarker this was a fairly extensive changeset and it came out pretty clean. Nice work. Some nitpicks below, but all are non-blocking. Feel free to push back on them.

src/phoebus_utils/adiabats.hpp
AstroBarker: done

AstroBarker:
Thanks for the feedback @Yurlungur -- all comments should be addressed.

inputs/torus.pin
```diff
@@ -137,3 +147,10 @@ rin = 6.0
 rmax = 12.0
 n_inside_horizon = 5
 magnetized = true
+entropy = 4
+
+# For tabulated EoS
+<units>
+scale_free = false
+geom_mass_msun = 2.8
+fluid_density_cgs = 1.e11
```
Yurlungur, line 156: This probably will cause tests to fail since I think they're based on ideal gas torus. Maybe we need two input decks?

Yurlungur:
Will merge as soon as tests pass. :+1:

inputs/torus.pin
AstroBarker: Good catch. I meant to comment that out. I think it's probably fine to leave it as 
```
#<units>
#scale_free = false
...
```
What do you think?

Yurlungur: yeah it's fine to just comment out the nuclear EOS stuff. Better in fact, so it's easy to see how to swap between modes

AstroBarker: For that matter I meant to do the same with the `<eos>` block.

PR #160, opened 2023-03-29
WIP: update some python tools array idnexing

AstroBarker:
At some point in the past the output format for some geometric quantities changed. A few accommodating changes are required, e.g., `x = data.coord[:,:,:,:,1]` -> `x = data.coord[:,1,:,:,:]`

I've made changes in the `plot_snap*.py` files for getting coordinates. Is there anywhere else a similar change would need to be made? I don't think so but I also don't remember if other geometric quantities changed output format. 

For `plot_snap2d.py` I've changed it to use `g.n.coords` for coordinates to avoid gaps at meshblock boundaries.
I also added optional support for [cmocean](https://matplotlib.org/cmocean/) because they have some great perceptually uniform colormaps that I like to use.

### DIFF
--- a/scripts/python/plot_snap2d.py
+++ b/scripts/python/plot_snap2d.py
@@ -20,9 +20,14 @@ x,y coordinates.
 from __future__ import print_function
 from argparse import ArgumentParser
 import numpy as np
+import warnings
 
 import matplotlib
 matplotlib.use('Agg')
+try:
+    import cmocean
+except:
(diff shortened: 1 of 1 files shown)


Yurlungur (approved):

Yurlungur:
@AstroBarker let me know when you want to merge this

AstroBarker:
> @AstroBarker let me know when you want to merge this

@Yurlungur I think this is fine

PR #161, opened 2023-03-30
Fix MC issue

brryan:
## PR Summary

We were calling `LinearInterpLog` inside the `while ()` of a `do {} while ()` loop in the Monte Carlo sampling, and for some reason on device (power9 for me) this generates a segfault when the `dNdlnu` `ParArrayND<Real>` is passed to that inline function, either as a copy or a const reference. Don't know why that is, but the simple fix is to write out the interpolation routine in the actual `par_for`. 

This runs the `thincooling.pin` problem to completion though I didn't check the output for accuracy.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/radiation/monte_carlo.cpp
+++ b/src/radiation/monte_carlo.cpp
@@ -278,21 +278,21 @@ TaskStatus MonteCarloSourceParticles(MeshBlock *pmb, MeshBlockData<Real> *rc,
             z(m) = minx_k + (k - kb.s + 0.5) * dx_k;
 
             // Sample energy and set weight
-            Real nu;
+            Real lnu;
             int counter = 0;
             Real prob;
             do {
-              nu = exp(rng_gen.drand() * (lnu_max - lnu_min) + lnu_min);
+              lnu = rng_gen.drand() * (lnu_max - lnu_min) + lnu_min;
               counter++;
               PARTHENON_REQUIRE(counter < 100000,
                                 "Inefficient or impossible frequency sampling!");
-              Real lnu = log(nu);
               Real dn = (lnu - lnu_min) / dlnu;
               int n = static_cast<int>(dn);
               dn = dn - n;
               prob = (1. - dn) * dNdlnu(n, sidx, k, j, i) +
                      dn * dNdlnu(n + 1, sidx, k, j, i);
             } while (rng_gen.drand() > prob);
+            Real nu = exp(lnu);
 
             weight(m) = GetWeight(wgtC / wgtCfac, nu);
 
--- a/src/radiation/monte_carlo.cpp
+++ b/src/radiation/monte_carlo.cpp
@@ -290,9 +290,9 @@ TaskStatus MonteCarloSourceParticles(MeshBlock *pmb, MeshBlockData<Real> *rc,
               Real dn = (lnu - lnu_min) / dlnu;
               int n = static_cast<int>(dn);
               dn = dn - n;
-              prob = (1. - dn) * dNdlnu(n, sidx, k, j, i) + dn * dNdlnu(n + 1, sidx, k, j, i);
+              prob = (1. - dn) * dNdlnu(n, sidx, k, j, i) +
+                     dn * dNdlnu(n + 1, sidx, k, j, i);
             } while (rng_gen.drand() > prob);
-
 
             weight(m) = GetWeight(wgtC / wgtCfac, nu);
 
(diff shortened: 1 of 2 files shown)


src/radiation/monte_carlo.cpp
```diff
@@ -259,12 +259,12 @@ TaskStatus MonteCarloSourceParticles(MeshBlock *pmb, MeshBlockData<Real> *rc,
           Geometry::Tetrads Tetrads(Ucon, Gcov);
           Real detG = geom.DetG(CellLocation::Cent, k, j, i);
           int dNs = v(iNs + sidx, k, j, i);
+          auto rng_gen = rng_pool.get_state();
 
           // Loop over particles to create in this zone
-          for (int n = 0; n < static_cast<int>(dNs); n++) {
```
brryan, line 264: Meaningless cast.

src/radiation/monte_carlo.cpp
```diff
@@ -259,12 +259,12 @@ TaskStatus MonteCarloSourceParticles(MeshBlock *pmb, MeshBlockData<Real> *rc,
           Geometry::Tetrads Tetrads(Ucon, Gcov);
           Real detG = geom.DetG(CellLocation::Cent, k, j, i);
           int dNs = v(iNs + sidx, k, j, i);
+          auto rng_gen = rng_pool.get_state();
```
brryan, line 262: Moved out of inner loop for efficiency.

src/radiation/monte_carlo.cpp
```diff
@@ -280,13 +280,19 @@ TaskStatus MonteCarloSourceParticles(MeshBlock *pmb, MeshBlockData<Real> *rc,
             // Sample energy and set weight
             Real nu;
             int counter = 0;
+            Real prob;
```
brryan, line 283: I wish I could do `Real prob = ` inside the `while {}` loop.

brryan (commented):
@AstroBarker This should fix the issue you were encountering. Let me know if it doesn't, thanks for telling me about the problem. If you are happy with this change feel free to go ahead and merge.

src/radiation/monte_carlo.cpp
```diff
@@ -280,13 +280,19 @@ TaskStatus MonteCarloSourceParticles(MeshBlock *pmb, MeshBlockData<Real> *rc,
...
             do {
               nu = exp(rng_gen.drand() * (lnu_max - lnu_min) + lnu_min);
```
Yurlungur, line 285: For efficiency, sample `lnu` here---don't take the exponent. Then you can skip the `lnu = log(nu)` line below. Then after the rejection sample in the while loop, compute `nu = exp(lnu)`

src/radiation/radiation.hpp
```diff
@@ -64,8 +64,9 @@ constexpr RadiationType species[MaxNumRadiationSpecies] = {
     RadiationType::NU_ELECTRON, RadiationType::NU_ELECTRON_ANTI, RadiationType::NU_HEAVY};
 
 KOKKOS_INLINE_FUNCTION
-Real LogLinearInterp(Real x, int sidx, int k, int j, int i, ParArrayND<Real> table,
+Real LogLinearInterp(Real x, int sidx, int k, int j, int i, const ParArrayND<Real> &table,
                      Real lx_min, Real dlx) {
+  PARTHENON_FAIL("This function cannot currently be called on device for some reason!");
```
Yurlungur, line 69: 🤷

Yurlungur (approved):
Thanks for fixing this @brryan . Minor performance and simplicity enhancement below: we're taking potentially `niter` too many logs/exponents in our rejection sampling loop, because we can just sample the log frequency for the rejection sampling and then compute the exponential after we find an acceptable sample.

src/radiation/monte_carlo.cpp
Yurlungur: Big Mood

AstroBarker (approved):
Thanks for this @brryan ! This runs thin cooling for me on the V100s. I do agree with @Yurlungur that it would be nice to do less log/exp calls. I'll work on verifying the output of the thin cooling problem, but I don't think that's blocking.

src/radiation/monte_carlo.cpp
brryan: Yeah good point -- fixed.

brryan:
OK I fixed the `exp/log` inefficiency -- good point noticing that. I think that inefficiency goes back almost 15 years lol. 

This passes the `thincooling.py` test so I assume everything still works. Merging now.

PR #162, opened 2023-04-05
Add LeptonSign functionality

AstroBarker:
This PR adds a function `LeptonSign` to `radiation.hpp` to return the lepton number based on radiation type. This has been incorporated into `monte_carlo` and `cooling_function` for the `Gye` updates.

`wgtC` is scaled by the time unit so that it can be compared to values computed in nubhlight (this was needed for the `leptoneq` problem). It's a temporary fix until it can be dynamically computed. 

I also modified the thin cooling setup to set the electron fraction based on the radiation type.

image
Output from the `leptoneq` problem (color axis is 100(ye - 0.225)).


Yurlungur (approved):
LGTM thanks @AstroBarker !

AstroBarker:
Regression tests failed, presumably due to changes to `wgtC`. I have undid those changes, since `wgtC` requires some hand-holding at the moment anyway.

PR #164, opened 2023-04-12
Fixes to monte carlo transport

AstroBarker:
This PR provides two fixes to transport. 
First, #162 adding lepton sign functionality had the lepton number for electrons and their antis backwards.. That is fixed. There are accompanying fixes in the emission updates in `monte_carlo` and `cooling_function`.

Secondly, there was an erroneous 4pi in the `alphanu` term for absorption. That has been removed. Below I show the fornax cooling (first) and thermalization (second) comparisons done with this PR. 

image
image


brryan (approved):
LGTM, thanks for catching the `4pi` bug. We will want to use your `LeptonSign` method in the radiation moments source term code as well but that needs a bit more work to loop `Ye` through fully so I can take a look at that later.

PR #165, opened 2023-04-19
Maintain sanity (or madness)

brryan:
## PR Summary

Long-duration simulations of Fishbone torii tend to accumulate large amounts of magnetic flux on the black hole (MAD state). While this may suggest that MADs are ubiquitous in nature, it would be nice to be able to simulate accretion onto weakly magnetized black holes (SANEs) for long times. This PR adds a runtime option for setting a specific `phi = Phi / sqrt(Mdot)` amd some timescale on which the code should maintain that value. Phi is maintained by adding a (hopefully small) amount of divergence-free nearly vertical field lines to the black hole every timestep. 

This PR also fixes a bug in the history flux evaluators, where the block index wasn't being passed in to the geometry calls. I think this should have been caught in the geometry infrastructure (it is caught with analytic but not cached geometry) but that is a separate issue (#171). The geometry worked correctly for single-meshblock-per-MPI-rank calculations.

Here is an example 256^2 2D torus run with no sanity maintenance (`torus.hst`) and the sanitized version with a target `phi = 1` (indicated by the dashed line in the phi plot) and a characteristic time for changing the net field `tau = 500M`:

image

The early hiccup in phi with the sanitized run is probably due to mdot being all over the place there. I also included a timer so that the phi changing code can be disabled for some initial coordinate time.

Restarts also look fine (here, a bit after 500M):
image

## PR Checklist

- [x] Explain what you did.
- [x] Use the Parthenon task-based reduction machinery.
- [ ] Do initialization only once. 
- [x] Ensure restarts work with flux enforcement logic.


src/phoebus_driver.cpp
```diff
@@ -351,6 +352,28 @@ TaskCollection PhoebusDriver::RungeKuttaStage(const int stage) {
...
+  }
+  // This is a bad pattern. Having a per-mesh data p2c would help.
+  TaskRegion &async_region_4 = tc.AddRegion(num_independent_task_lists);
+  for (int i = 0; i < blocks.size(); i++) {
+    auto &pmb = blocks[i];
+    auto &tl = async_region_4[i];
+    auto &sc = pmb->meshblock_data.Get(stage_name[stage]);
+
+    auto p2c = tl.AddTask(none, fluid::PrimitiveToConserved, sc.get());
+  }
```
brryan, line 375: This new set of regions is good motivation for getting the rest of the hydro to be per-meshdata I think

src/fixup/fixup.cpp
```diff
@@ -71,6 +73,23 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   bool enable_mhd_floors = pin->GetOrAddBoolean("fixup", "enable_mhd_floors", false);
   bool enable_rad_floors = pin->GetOrAddBoolean("fixup", "enable_rad_floors", false);
   bool enable_rad_ceilings = pin->GetOrAddBoolean("fixup", "enable_rad_ceilings", false);
+  bool enable_phi_enforcement =
+      pin->GetOrAddBoolean("fixup", "enable_phi_enforcement", false);
+  params.Add("enable_phi_enforcement", enable_phi_enforcement);
+  Real enforced_phi = pin->GetOrAddReal("fixup", "enforced_phi", 0.);
```
Yurlungur, line 79: The default here probably shouldn't be zero.

src/fixup/fixup.cpp
```diff
@@ -645,6 +664,156 @@ TaskStatus ApplyFloorsImpl(T *rc, IndexDomain domain = IndexDomain::entire) {
...
+  auto tr = Geometry::McKinneyGammieRyan(derefine_poles, h, xt, alpha, x0, smooth);
+
+  if (enable_phi_enforcement) {
+    const Real enforced_phi = fix_pkg->Param<Real>("enforced_phi");
+    const Real enforced_phi_timescale = fix_pkg->Param<Real>("enforced_phi_timescale");
+    const Real enforced_phi_cadence = fix_pkg->Param<Real>("enforced_phi_cadence");
+    const Real enforced_phi_start_time = fix_pkg->Param<Real>("enforced_phi_start_time");
+    if (t > enforced_phi_start_time) {
+
+      // This only needs to be done at initialization
```
Yurlungur, line 708: Could hide it behind a static bool that gets set to true after it's been done once?

src/fixup/fixup.cpp
```diff
@@ -645,6 +664,156 @@ TaskStatus ApplyFloorsImpl(T *rc, IndexDomain domain = IndexDomain::entire) {
...
+      if (last_stage && t >= next_dphi_dt_update_time) {
+        next_dphi_dt_update_time += enforced_phi_cadence;
+
+        // Record mdot
+        const Real mdot_proc = History::ReduceMassAccretionRate(md);
+        const Real mdot = reduction::Sum(mdot_proc);
+
+        // Measure phi
+        const Real phi_proc = History::ReduceMagneticFluxPhi(md);
+        const Real phi = reduction::Sum(phi_proc) / std::sqrt(mdot);
```
Yurlungur, line 742: Note of caution here---if we have more than one `MeshData` object per MPI rank, this might hang? Probably better to use task-based reducers... or enforce somewhere that this is the only `MeshData` object on a given MPI rank for this task region

src/fixup/fixup.cpp
```diff
@@ -645,6 +664,156 @@ TaskStatus ApplyFloorsImpl(T *rc, IndexDomain domain = IndexDomain::entire) {
...
+                   (2.0 * coords.CellWidthFA(X1DIR, k, j, i) * gammadet));
+
+              // Used by ReduceMagneticFluxPhi
+              pack(b, cblo, k, j, i) = pack(b, pblo, k, j, i) * gammadet;
+              pack(b, cblo + 1, k, j, i) = pack(b, pblo + 1, k, j, i) * gammadet;
+            });
+
+        // Measure change in phi
+        const Real fiducial_phi_proc = History::ReduceMagneticFluxPhi(md);
+        const Real fiducial_phi = reduction::Sum(fiducial_phi_proc) / std::sqrt(mdot);
```
Yurlungur, line 769: Same as above

src/fixup/fixup.cpp
```diff
@@ -645,6 +664,156 @@ TaskStatus ApplyFloorsImpl(T *rc, IndexDomain domain = IndexDomain::entire) {
...
+
+            pack(b, pblo, k, j, i) +=
+                phi_factor * dt *
+                (-(A(b, j, i) - A(b, j + 1, i) + A(b, j, i + 1) - A(b, j + 1, i + 1)) /
+                 (2.0 * coords.CellWidthFA(X2DIR, k, j, i) * gammadet));
+            pack(b, pblo + 1, k, j, i) +=
+                phi_factor * dt *
+                ((A(b, j, i) + A(b, j + 1, i) - A(b, j, i + 1) - A(b, j + 1, i + 1)) /
+                 (2.0 * coords.CellWidthFA(X1DIR, k, j, i) * gammadet));
+          });
```
Yurlungur, line 810: Why can't we combine these two kernels? `A` doesn't change so feels like fiducial B-field and update can be done in the same loop?

src/phoebus_driver.cpp
```diff
@@ -351,6 +352,28 @@ TaskCollection PhoebusDriver::RungeKuttaStage(const int stage) {
     }
   }
 
+  // Extra per-step user work
+  TaskRegion &sync_region_5 = tc.AddRegion(num_partitions);
```
Yurlungur, line 356: You can have as many partitions or as few as you want on a per-region basis. Just set this to 1. Then you have to create a new meshdata object specifically for this task region that just owns all blocks.

src/phoebus_driver.cpp
```diff
@@ -351,6 +352,28 @@ TaskCollection PhoebusDriver::RungeKuttaStage(const int stage) {
...
+  PARTHENON_REQUIRE(num_partitions == 1,
+                    "Reductions don't work for multiple partitions?");
```
Yurlungur, line 358: This is what I was saying above, it's because you have multiple `MPI_AllReduce`s per rank. If you use the Parthenon reduction machinery this problem goes way. Or you can make sure this task region *only* gets called with 1 partition.

Yurlungur (commented):
Good idea. Some comments below. Nothing really blocking, but I think we shouldn't demand that `num_partitions=1` across the whole code if we can help it, as that limits what we can do with load balancing (i.e., with monte carlo) and with AMR (when we get there)

src/fixup/fixup.cpp
brryan: I think 0 is ok -- no net flux is a valid state, and I don't think there are really any special numbers > 0 (apart from the MAD limit but I don't think this machinery will work well there or above anyway because it'll produce extra flux that will hop off the black hole into the disk, and then machinery will generate more flux, etc. etc.)

src/fixup/fixup.cpp
brryan: Yeah I need to not be lazy and just use the parthenon reductions. I'll make that change...

src/fixup/fixup.cpp
brryan: Yeah that should be fine for restarts (actually I need to switch some of my existing static bools to mutable params (I think) so restarts work with them)

src/fixup/fixup.cpp
brryan: The former kernel is only being run when dphi/dt is being recalculated (typically only once every many timesteps). I think the former kernel can be bundled into the initialization step anyway and only done once.

src/phoebus_driver.cpp
brryan: Yep, will fix.

src/phoebus_driver.cpp
brryan: Good workaround, but yeah I should just do the right thing

src/fixup/fixup.cpp
brryan: Thinking more I'm not sure the `static bool` trick will work here if we have multiple regions per MPI rank.

brryan: To do this fully correctly (once *per restart*) the thing to do is probably create a mutable parameter for whether the normalization has been evaluated, defaults to false, and is set to true when normalization is evaluated. Then a similar mutable param that stores the normalization parameter. 

Then one *should* also store `dphi_dt` and `next_dphi_dt_update_time` in a parthenon Variable that is carried across restarts. 

Neither of these is necessary to make the thing actually work -- I think the hiccup in `dphi_dt` evaluation cadence during restarts is probably not significant...

src/fixup/fixup.cpp
brryan: I did it! Though maybe I should put the `enable_phi_enforcement` check in the task list itself for brevity...

src/fixup/fixup.cpp
brryan: Done

scripts/python/phoedf.py
```diff
@@ -80,11 +81,37 @@ def flatten_indices(mu, nu):
                 self.flatgcov[:, flatten_indices(1, 1), :, :, :] = 1.0
                 self.flatgcov[:, flatten_indices(2, 2), :, :, :] = 1.0
                 self.flatgcov[:, flatten_indices(3, 3), :, :, :] = 1.0
+            else:
+                assert (
+                    False
+                ), "Geometry not provided in dump file but also not Minkowski!"
         self.gcov = np.zeros([self.NumBlocks, 4, 4, self.Nx3, self.Nx2, self.Nx1])
+
+        # Account for included ghost zones in flattened geometry data
```
brryan, line 90: With the new variable output logic in parthenon, gcov, which is technically not a proper cell-centered variable because it can suppress the X3 dimension in 3D, no longer had its ghost cells automatically removed along with the proper cell-centered variables when this is requested. So here we remove the ghost zones if necessary to match the rest of the data.

src/phoebus_utils/reduction.hpp
```diff
@@ -32,10 +32,17 @@ Real inline Min(const Real &x) {
   return xmin;
 }
 
+Real inline Sum(const Real &x) {
```
brryan, line 35: Not actually used now but I think we can leave this in in case the need arises in a problem generator file at some point

tst/regression/regression_test.py
```diff
@@ -216,6 +246,17 @@ def gold_comparison(
     compression_factor=1,
     tolerance=1.0e-5,
 ):
+    problem = read_input_value("phoebus", "problem", input_file)
```
brryan, line 249: It was difficult to understand what tests weren't working from the CI without this.

brryan:
OK this is ready for re-review @Yurlungur. The function used to paint on B field lines is still unnecessarily re-evaluated every so often but this cost should always be small so I didn't bother to rewrite everything to work with restarts for this pretty specific feature. If you think this is too ugly or problematic let me know and I can revisit. Otherwise this feature seems to do what it's supposed to.

src/fixup/fixup.cpp
Yurlungur: I'm just thinking is 0 the flux you want to preserve by default?

Yurlungur (approved):

src/fixup/fixup.cpp
brryan: I think zero is fine, its just saying "as small a flux as possible given the flux tuning parameters" which is sort of the only valid "special" value. This flux enforcement is off by default of course so I imagine it will only be used when someone has a specific flux to enforce in mind. 

The other option would be to not provide a default value at all for `enforced_phi` which I am fine with if you think that's better/safer.

Yurlungur: I would rather not provide a default and just die if it's on without a value set.

src/phoebus_driver.cpp
brryan: Yes good catch thanks, it's removed.

src/fixup/fixup.cpp
brryan: Yeah I agree -- default value removed.

PR #166, opened 2023-04-19
Hyperexponential coordinates

brryan:
## PR Summary

An alternative (going back to ~Tchekhovskoy+ 2011) to having an outer boundary near a disk is to use hyperexponential coordinates such that, beyond some break radius, zones grow radially in size very rapidly with radius.

This PR implements the hyperexponential coordinates based on how they are implemented in KHARMA.

## PR Checklist

- [x] Explain what you did.
- [x] Do at least some rudimentary testing.
- [x] Make sure this resolves to the `flatgcov` treatment in #165

### DIFF
--- a/scripts/bash/format.sh
+++ b/scripts/bash/format.sh
@@ -51,6 +51,7 @@ if ! command -v ${PFM} &> /dev/null; then
 else
     PFM=$(command -v ${PFM})
     echo "black Python formatter found: ${PFM}"
+    echo "black version: $(${PFM} --version)"
 fi
 
 echo "Formatting Python files..."
@@ -59,6 +60,6 @@ for f in $(git grep --untracked -ail res -- :/*.py); do
     if [ ${VERBOSE} -ge 1 ]; then
        echo ${f}
     fi
-    ${PFM} ${REPO}/${f}
+    ${PFM} -q ${REPO}/${f}
 done
 echo "...Done"
--- a/scripts/python/phoedf.py
+++ b/scripts/python/phoedf.py
@@ -75,7 +75,7 @@ class phoedf(phdf.phdf):
     self.gcov = np.zeros([self.NumBlocks, 4, 4, self.Nx3, self.Nx2, self.Nx1])
     for mu in range(4):
       for nu in range(4):
-        self.gcov[:,mu,nu,:,:,:] = self.flatgcov[:,flatten_indices(mu,nu),:,:,:]
+        self.gcov[:,mu,nu,:,:,:] = self.flatgcov[:,flatten_indices(mu,nu),:,4:-4,4:-4]
     del(self.flatgcov)
     self.gcon = np.zeros([self.NumBlocks, 4, 4, self.Nx3, self.Nx2, self.Nx1])
     for b in range(self.NumBlocks):
--- a/src/geometry/fmks.cpp
+++ b/src/geometry/fmks.cpp
@@ -46,6 +46,9 @@ void Initialize<FMKSMeshBlock>(ParameterInput *pin, StateDescriptor *geometry) {
   Real alpha = pin->GetOrAddReal("coordinates", "poly_alpha", 14);
   Real x0 = pin->GetReal("parthenon/mesh", "x1min");
   Real smooth = pin->GetOrAddReal("coordinates", "smooth", 0.5);
+  Real hexp_br = pin->GetOrAddReal("coordinates", "hexp_br", 1000.);
+  Real hexp_nsq = pin->GetOrAddReal("coordinates", "hexp_nsq", 1.0);
+  Real hexp_csq = pin->GetOrAddReal("coordinates", "hexp_csq", 4.0);
   Real a = pin->GetReal("geometry", "a");
   Real Rh = 1.0 + sqrt(1.0 - a * a);
   Real xh = log(Rh);
@@ -72,6 +75,9 @@ void Initialize<FMKSMeshBlock>(ParameterInput *pin, StateDescriptor *geometry) {
   x0 = std::min<Real>(x0, std::max<Real>(0., x1min - nghost * dx1));
   params.Add("x0", x0);
   params.Add("smooth", smooth);
+  params.Add("hexp_br", hexp_br);
+  params.Add("hexp_nsq", hexp_nsq);
+  params.Add("hexp_csq", hexp_csq);
   params.Add("Rh", Rh);
   params.Add("xh", xh);
   params.Add("r_isco_p", r_isco_p);
--- a/src/geometry/fmks.cpp
+++ b/src/geometry/fmks.cpp
@@ -46,7 +46,7 @@ void Initialize<FMKSMeshBlock>(ParameterInput *pin, StateDescriptor *geometry) {
   Real alpha = pin->GetOrAddReal("coordinates", "poly_alpha", 14);
   Real x0 = pin->GetReal("parthenon/mesh", "x1min");
   Real smooth = pin->GetOrAddReal("coordinates", "smooth", 0.5);
-  Real hexp_br = pin->GetOrAddReal("coordinates", "hexp_br", 1000.);
+  Real hexp_br = pin->GetOrAddReal("coordinates", "hexp_br", 1.e6);
   Real hexp_nsq = pin->GetOrAddReal("coordinates", "hexp_nsq", 4.0);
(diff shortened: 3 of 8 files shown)


src/geometry/mckinney_gammie_ryan.hpp
```diff
@@ -66,7 +72,10 @@ class McKinneyGammieRyan {
     C[1] = th;
     C[2] = X3;
 
-    const Real drdX1 = std::exp(X1);
+    // const Real drdX1 = std::exp(X1);
+    const Real hexp_dr = X1 - hexp_br_;
+    const Real drdX1 =
+        std::exp(X1 + (hexp_dr > 0.) * hexp_csq_ * std::pow(hexp_dr, hexp_nsq_));
```
Yurlungur, line 78: nice use of masking. I was worried there'd be an if statement.

Yurlungur (approved):
We should probably make sure this doesn't break anything and also add a test. Other than that, :+1: nice improvement.

I *would* like someday to fully compactify the domain and go all the way to physical infinity at $X1=1$ or something. But that can be a problem for future us.

src/geometry/mckinney_gammie_ryan.hpp
bprather: Is drdX1 supposed to be the derivative?  If so this is missing a term, right?  (vs [here](https://github.com/AFD-Illinois/kharma/blob/feature/multizone-dev/kharma/coordinates/coordinate_systems.hpp#L476))

Incidentally, if Phoebus uses numerical first derivatives/transformation matrices, why?

src/geometry/mckinney_gammie_ryan.hpp
```diff
@@ -191,6 +202,9 @@ class McKinneyGammieRyan {
   Real x0_ = 0; // start point of smooth region
   Real smooth_ = 0.5;
   Real norm_;
+  Real hexp_br_ = 1000.;
+  Real hexp_nsq_ = 1.;
```
bprather, line 206: As I discovered when I tried to run things with this value, nsq=1 is dangerous (it is the default in HARMPI and what was used in Tchekhovskoy '11 afaict though).
The first derivative of the function r(X1) is discontinuous for this value, and it exacerbates the problems with passing material over the break radius.  The effect may be much less if you use numerical first derivatives, as HARMPI does.

bprather (approved):
Code sharing!

src/geometry/mckinney_gammie_ryan.hpp
brryan: @Yurlungur the adroit use of masking is due to @bprather :) Does this actually save time compared to the if statement? It's still evaluating a conditional right? I could be wrong, I'm just curious.

@bprather Yes you're right I dropped a term, thanks for catching that. I'll fix. I'm not sure I understand the numerical transformation matrix concern, could you elaborate?

src/geometry/mckinney_gammie_ryan.hpp
brryan: Ah ok, I just pulled out the default values from kharma (as I interpreted them). Could you remind me what your preferred values are and I can update these defaults?

brryan:
> Code sharing!

Does copy and pasting count? Lol

src/geometry/mckinney_gammie_ryan.hpp
Yurlungur: @brryan  the masking vectorize whereas a conditional will not... although it's not clear this vectorizes anyway. (The geometry package as a while probably does not.)

bprather: It's not a concern, just trying to gauge.  I think in this case, numerical first+second derivatives would actually be more stable, because analytically speaking the derivative of HExp can change really quickly.  Then, a numerical second derivative of the sharp first derivative causes :sparkles: problems :sparkles:

In particular, d/dx of exp(x + pow(x,n)) is exp(x + pow(x, n))*(1 + n*exp(x + pow(x, n-1)), yeah? (Or in any case involves pow(x, n-1), I'm dashing this off).  However, if n=1, pow(x,0) is constant for *all* x, i.e., including x=0.  Thus, the derivative becomes straight-up mathematically discontinuous at x=x_br.  for n~=1, it's still pretty bad from a numerical-derivative perspective, and can lead to some gnarly connection coefficients.  And even for n=2 (which starts to *really* stretch the outer edge) it seems to have a noticeable effect, stemming I think from not preserving conn^mu_mu_nu = d/dx^nu gdet

bprather:
> > Code sharing!
> 
> Does copy and pasting count? Lol

You changed the names!

(Honestly I like your naming better, and hadn't considered adding HExp to FMKS since I'm using it so differently, so I'll probably steal it back even more literally)

src/geometry/mckinney_gammie_ryan.hpp
Yurlungur: as a general rule, Phoebus supports analytic *and* numerical derivatives. Likely when this code is run, the derivatives will be numerical if the code is run with cached geometry.

src/geometry/mckinney_gammie_ryan.hpp
Yurlungur: :grin:

brryan: Thanks -- updated to `hexp_nsq = 4`

brryan:
> We should probably make sure this doesn't break anything and also add a test. Other than that, 👍 nice improvement.

The original FMKS unit test was actually good for this -- the answer is unchanged in the region below the hyperexponential break radius. I also updated that test to sample a region outside the break radius.

scripts/python/phoedf.py
```diff
@@ -75,7 +75,7 @@ def flatten_indices(mu, nu):
     self.gcov = np.zeros([self.NumBlocks, 4, 4, self.Nx3, self.Nx2, self.Nx1])
     for mu in range(4):
       for nu in range(4):
-        self.gcov[:,mu,nu,:,:,:] = self.flatgcov[:,flatten_indices(mu,nu),:,:,:]
```
brryan, line 78: I'm not immediately sure why this change was necessary, I should figure this out

brryan: It was due to lack of ghost zone pruning for non-proper-variable data like `gcov` after the recent changes to parthenon output. This is properly fixed in #165

Yurlungur:
Is this ready to go in @brryan ?

brryan:
@Yurlungur This is ready to go in, but I think it would be easiest to merge #165 first when everyone thinks thats ready, then I can deal with the merge conflicts and undoing the less-good fix to `gcov` ghost cell pruning in `phoedf.py` that is currently in this PR if thats OK with you

src/geometry/fmks.cpp
```diff
@@ -46,6 +46,9 @@ void Initialize<FMKSMeshBlock>(ParameterInput *pin, StateDescriptor *geometry) {
   Real alpha = pin->GetOrAddReal("coordinates", "poly_alpha", 14);
   Real x0 = pin->GetReal("parthenon/mesh", "x1min");
   Real smooth = pin->GetOrAddReal("coordinates", "smooth", 0.5);
+  Real hexp_br = pin->GetOrAddReal("coordinates", "hexp_br", 1.e6);
```
brryan, line 49: Hyperexponential coordinates are on by default... I only see this value being a problem for maybe very large cutout simulations that don't expect hyperexponential coordinates. Maybe this default value should be `std::numeric_limits<Real>::max()`?

scripts/bash/format.sh
```diff
@@ -59,6 +60,6 @@ for f in $(git grep --untracked -ail res -- :/*.py); do
     if [ ${VERBOSE} -ge 1 ]; then
        echo ${f}
     fi
-    ${PFM} ${REPO}/${f}
+    ${PFM} -q ${REPO}/${f}
```
brryan, line 63: Too many cake emojis for unchanged files

src/geometry/fmks.cpp
Yurlungur: I would prefer `std::numeric_limits<Real>::max()` yeah

brryan: `numeric_limits::max()` caused problems so I just used `1.e30` with a check that that's smaller than `Real`.

PR #168, opened 2023-04-20
WIP: Number Moments

lroberts36:
## PR Summary
This PR implements the radiation number moment evolution equations. Previously, we just evolved the conserved energy and momentum of the radiation field for each species when using the moments. Now, if the parameter `radiation/do_number_evolution` is set, the conserved number density and a corresponding number flux are also evolved for each species. See the Phoebus notes for a description of the equations that are being implemented. 

Things that still need to get done: 

- [ ] Update `FixFluxes` to operate on both number and energy moments (this requires splitting into separate functions for hydro and radiation, at least to minimize code duplication). 
- [ ] Update `ApplyFloors` to operate on both number and energy moments.
- [ ] Add `MomentRedshiftingSource` task, which includes the extra RHS terms the number flux equation picks up due to redshifting. This requires calculating the expansion, acceleration, and shear of the fluid 4-velocity field. The expansion and acceleration are also need to be included in the asymptotic flux expression. 
- [ ] Update opacity function to get equilibrium number density. Does singularity provide this? 
- [ ] Update source functions to allow for multiple species and for evolution of $Y_e$. This is probably the biggest chunk of work. 
- [ ] Run old tests to make sure changes don't break anything. 
- [ ] Set up and run $(T, Y_e)$ and redshifting tests

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/fixup/fixup_radc2p.cpp
+++ b/src/fixup/fixup_radc2p.cpp
@@ -41,14 +41,11 @@ using robust::ratio;
 
 namespace fixup {
 
-template <typename T, class CLOSURE>
+template <typename T, class CLOSURE, radiation::RadEnergyMoment EMOMENT>
 TaskStatus RadConservedToPrimitiveFixupImpl(T *rc) {
   namespace p = fluid_prim;
   namespace c = fluid_cons;
   namespace impl = internal_variables;
-  namespace ir = radmoment_internal;
-  namespace pr = radmoment_prim;
-  namespace cr = radmoment_cons;
 
   auto *pmb = rc->GetParentPointer().get();
   IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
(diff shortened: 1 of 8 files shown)


brryan:
```
Update opacity function to get equilibrium number density. Does singularity provide this?
```

`singularity-opac` provides `B = \int \frac{1}{h \nu} B_{\nu} d\Omega d\nu` as `ThermalNumberDistributionOfT`; this is what you need right? Although that function will need to be wrapped by `phoebus`'s `Opacities` class in `src/microphysics/opac_phoebus/opac_phoebus.hpp` (should work exactly like `ThermalDistributionOfTNu` without the `nu` argument). The frequency-dependent version of number distribution can of course just be calculated as `ThermalDistributionOfTNu / (h nu)`

lroberts36:
@brryan: Yep, that is exactly what I need. Thanks!

PR #172, opened 2023-05-09
WIP: Add k_mu source terms

AstroBarker:
This PR adds calculations for `k_mu` source terms (`radiation/geodesics.hpp`) for arbitrary metric/geometry as it was previously specialized to Minkowski. This was causing unphysical positions when running the torus problem. It now runs beyond the problem point and a longer run in in the queue. The `leptoneq` problem seems reasonable.

Also removes a print statement from `plot_snap2d` that I left from before.

TODO:
 - [ ] There are redundant calculations (in Minowski all the sources are 0, in FMKS only 2 are non zero) and we are doing more integration than necessary. Performant way to specialize? Template on geometry?
 - [ ] Write a simple test for particles around a Kerr black hole to say for sure that the Push is doing what we want.

### DIFF
--- a/scripts/python/plot_snap2d.py
+++ b/scripts/python/plot_snap2d.py
@@ -70,7 +70,6 @@ def plot_dump(
     coordname = "g.n.coord"
     coord = data.Get(coordname, False)
     z = coord[:, 3, :, nG - 1 : -1 - nG, nG - 1 : -1 - nG]
-    print(np.shape(z))
     y = coord[:, 2, :, nG - 1 : -1 - nG, nG - 1 : -1 - nG]
     x = coord[:, 1, :, nG - 1 : -1 - nG, nG - 1 : -1 - nG]
 
@@ -100,7 +99,7 @@ def plot_dump(
     fig = plt.figure()
     p = fig.add_subplot(111, aspect=1)
     for i in range(NB):
-        val = q[i, 0, :, :]
+        val = 100.0 * (q[i, 0, :, :] - 0.225)
         if len(val.shape) > 2:
             print("WARNING plotting the 0th index of multidimensional variable!")
             val = val[:, :, 0]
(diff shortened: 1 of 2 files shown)


scripts/python/plot_snap2d.py
```diff
@@ -100,7 +99,7 @@ def plot_dump(
     fig = plt.figure()
     p = fig.add_subplot(111, aspect=1)
     for i in range(NB):
-        val = q[i, 0, :, :]
+        val = 100.0 * (q[i, 0, :, :] - 0.225)
```
brryan, line 102: Leftover code to be undone?

src/radiation/geodesics.hpp
```diff
@@ -39,7 +39,34 @@ KOKKOS_INLINE_FUNCTION
...
-  SPACETIMELOOP(mu) { source[mu] = 0.; }
+  Real Kcov[NDFULL] = {Kcov0, Kcov1, Kcov2, Kcov3};
+  Real Xm[NDFULL], Xp[NDFULL];
+  Real Gconm[NDFULL][NDFULL], Gconp[NDFULL][NDFULL];
+  Real dG;
+
+  constexpr Real DELTA = 1.0e-6;
+
+  SPACETIMELOOP(mu) {
+    source[mu] = 0.;
```
brryan, line 50: Maybe just add a `TODO()` here pointing out that we can come back and exploit spacetime symmetries to reduce the compute burden later?

brryan (commented):
Great work! I can provide comparison data for the particle pushes soon. A lot of this can be done in a unit test, but ultimately I think it would be good to compare particles that have been pushed through the driver, to ensure that driver stages are interacting appropriately with the particle pushes. In that case a purpose-specific `pgen` that creates a few initial particles and then outputs particle positions with particle IO may be the right choice.

scripts/python/plot_snap2d.py
AstroBarker: ah yikes yes

src/radiation/geodesics.hpp
AstroBarker: done

Yurlungur (approved):
Nice work! As @brryan suggested, this is probably something we should be integrating more tightly with the geometry package, so that nothing here needs to be recalculated when there are symmetries. But that can be a future improvement.

PR #173, opened 2023-05-15
WIP: implement relativistic standing accretion shock test problem

carlnotsagan:
## PR Summary

We want to implement the relativistic standing accretion shock test case to explore high resolution turbulent convection. 

References: 

https://iopscience.iop.org/article/10.3847/0004-637X/820/1/76/pdf
Eq 4. + ...

https://iopscience.iop.org/article/10.1088/2041-8205/813/1/L6/pdf
https://iopscience.iop.org/article/10.1086/508443/pdf

## Issues:

- fails to bracket root when executing driver, setup executing completes however and values seem reasonable


src/pgen/standing_accretion_shock.cpp
```diff
@@ -0,0 +1,248 @@
...
+          const Real W0 = 1. / sqrt(1. - std::pow(vr0, 2));
+          Real rho0 = Mdot / (4. * M_PI * std::pow(r, 2) * W0 * std::abs(vr0));
+
+          Real alphasq = 1. - (2. / r);
+          Real psi = alphasq * ((gamma - 1.) / gamma) * ((W0 - 1.) / W0);
+          Real vr1 = (vr0 + std::sqrt(vr0 * vr0 - 4. * psi)) / 2.;
+          Real rho1 = rho0 * W0 * (vr0 / vr1);
+
+          v(irho, k, j, i) = rho1;
+          v(ieng, k, j, i) = (W0 - 1. + epsND * (gamma - 1.)) / (gamma);
```
Yurlungur, line 150: `v(ieng, k, j, i)` is internal energy by volume and formula from Radice is specific. Need to multiply by rho.

Yurlungur: Need to solve Radice's shock equations for EOS without assuming gamma

carlnotsagan: So the issue I have run into is needing to choose a variable that we know to solve the rest. I settled on using the equation for velocity in the preshock region assuming ideal gas: const Real vr1 = (vr0 + std::sqrt(vr0 * vr0 - 4. * psi)) / 2.;

Then the density follows from the Rankine–Hugoniot jump conditions (no ideal gas assumption).

Then I solve the remaining equations without ideal gas assumption using vr1,vr0,rho0,rho1.

PR #183, opened 2023-10-26
Small history fix for FMKS without MHD

AstroBarker:
When I wrote the history stuff to compute e.g., mass accretion rate and magnetic flux, I checked that we did indeed `do_hydro` .... but not that we are doing MHD. This caused the torus problem without MHD to crash. This should fix that.

### DIFF
--- a/src/geometry/geometry.cpp
+++ b/src/geometry/geometry.cpp
@@ -60,7 +60,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   geometry->AddField(geometric_variables::node_coords, gcoord_node);
 
   // Reductions
-  const bool do_mhd = pin->GetBoolean("fluid", "mhd");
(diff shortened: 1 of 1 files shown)


src/geometry/geometry.cpp
```diff
@@ -60,8 +60,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   geometry->AddField(geometric_variables::node_coords, gcoord_node);
 
   // Reductions
-  const bool do_hydro = pin->GetBoolean("physics", "hydro");
-  if (params.hasKey("xh") && do_hydro) {
+  const bool do_mhd = pin->GetOrAddBoolean("fluid", "mhd", false);
+  if (params.hasKey("xh") && do_mhd) {
```
Yurlungur, line 64: Should you check both hydro and mhd? Just to be safe?

Yurlungur (approved):

PR #184, opened 2023-11-07
Inconsistent behavior between CPU and GPU for liebendorf deleptonization and cooling

carlnotsagan:
Issue:
--
when using `cooling_function` with SASW/homologous problem. returns error: 

`:0: : block: [3,0,0], thread: [0,0,0] Assertion `View bounds error of view p.density` failed.`

after first step

this backtraces to:

[the cooling function](https://github.com/lanl/phoebus/blob/55e6c2656b05b574766e804e7c4667428e0f7b47/src/radiation/cooling_function.cpp#L84C1-L84C42)

Host function was being called from device function leading to issues on GPU. 

---

Steps Taken: 
--
- moved units calls outside of device kernels and saved as reals for call within kernel


src/phoebus_driver.cpp
```diff
@@ -685,7 +685,7 @@ TaskListStatus PhoebusDriver::RadiationPostStep() {
     parthenon::AllReduce<bool> *pdo_gain_reducer;
     bool do_lightbulb = rad->Param<bool>("do_lightbulb");
     if (do_lightbulb) {
-      pdo_gain_reducer = rad->MutableParam<parthenon::AllReduce<bool>>("do_gain_reducer");
+      pdo_gain_reducer = rad->MutableParam<parthenon::AllReduce<bool>("do_gain_reducer");
```
Yurlungur, line 688: This isn't a typo. Notice that `MutableParam` contains a `<` and then `AllReduce` does. So you need two `>` at the end.

carlnotsagan: oh wow. facepalm. ok well i will keep this open if i find something else relating to the density error. thanks

Yurlungur: :+1: sounds good

src/radiation/cooling_function.cpp
```diff
@@ -175,7 +175,7 @@ TaskStatus CoolingFunctionCalculateFourForce(MeshBlockData<Real> *rc, const doub
     singularity::StellarCollapse eos_sc =
         eos.GetUnmodifiedObject().get<singularity::StellarCollapse>();
     const parthenon::AllReduce<bool> *pdo_gain_reducer =
-        rad->MutableParam<parthenon::AllReduce<bool>("do_gain_reducer");
+        rad->MutableParam < parthenon::AllReduce<bool>("do_gain_reducer");
```
mari2895: rad->MutableParam < parthenon::AllReduce<bool>>("do_gain_reducer");  maybe?

mari2895: missing second > here as well.

Yurlungur:
@carlnotsagan do you encounter the same issue when running lightbulb with @mari2895 's CCSN problem initialization? If not maybe it has to do with changes in the SASW branch? Can you share the code for SASW in a PR?

Yurlungur (approved):
@mari2895 can you briefly skim this to make sure you agree? Then I will merge.

carlnotsagan:
thanks for merging main, should be all good now

src/radiation/cooling_function.cpp
mari2895, line 1: Looks good to me

Yurlungur:
Looks like tests are still failing?

mari2895 (approved):

carlnotsagan:
looking into the fails. something to do with a reference to pmb

src/radiation/cooling_function.cpp
```diff
@@ -179,14 +189,24 @@ TaskStatus CoolingFunctionCalculateFourForce(MeshBlockData<Real> *rc, const doub
...
+
+    // Code to CGS
+    const Real density_conversion_factor = unit_conv.GetMassDensityCodeToCGS();
+    const Real temperature_conversion_factor = unit_conv.GetTemperatureCodeToCGS();
+    const Real length_conversion_factor = unit_conv.GetLengthCodeToCGS();
+
+    // CGS to code
+    const Real energy_conversion_factor = unit_conv.GetEnergyCGSToCode();
+    const Real mass_conversion_factor = unit_conv.GetMassCGSToCode();
+    const Real time_conversion_factor = unit_conv.GetTimeCGSToCode();
```
mari2895, line 201: These lines probably don't need to be here, because the same definitions are above in the same function.

Yurlungur (approved):

PR #186, opened 2023-11-15
fix typo in relative path of Cmake

carlnotsagan:
typo in relative path for Cmake file variant patch


Yurlungur:
This is not a typo---it depends on where you call `cmake`. However what's in there currently is a hack---there's a more robust way to do this, which I will try to thread through.

carlnotsagan:
ok, makes sense, ive changed the title of the pr to reflect that

Issue #187, opened 2023-12-02
Long term solution for gold files


AstroBarker:
As the title says. I am starting to spin up implementing the [Parthenon regression testing framework](https://parthenon-hpc-lab.github.io/parthenon/develop/src/tests.html#regression-tests) into Phoebus. It's basically done, but we are going to need a long term solution for gold files for accepted solutions. Parthenon bakes them into github releases -- are there size limits on that? i.e., would we run into issues far down the line? Thoughts?

brryan:
I strongly prefer not switching to the release based gold model used in
Parthenon. It requires manually syncing PRs with gold releases which I find
awkward and laborious to set up for the developer, and to review for the
reviewer. I believe golds should be in plaintext and included in PRs as
they are now. The compression option we have in Phoebus where only a sparse
sampling of the gold file can be included in the repository for large files
was supposed to make this workable long-term. Is there a looming problem
with what we have now in Phoebus?

Best

Ben

AstroBarker:
Ah I see. I  missed the process for creating the gold files -- I like this far more than the release model. Should've looked closer. As for looming problems, none. What are thoughts on wrapping the existing infrastructure into the Parthenon tooling for running the tests, so that they e.g., hook into the cmake testing?

Yurlungur:
I am concerned about the long-term viability of plaintext gold files but they've worked for us up until now, so I'm fine keeping that machinery in place. I do like the idea of hooking into Parthenon's testing machinery other than that.

jonahm-LANL:
The Parthenon testing machinery does support plaintext gold files as it just reads whatever you tell it to in the python script.

IMO catch and our regression tests serve different needs, and we’re already using catch for our unit tests.

Parthenon’s regression test suite is actually very similar to what we wrote forever ago for firelink/MOCMC. (And In fact the design is informed by that).

That said, I don’t feel strongly either way, except that it would be nice to be running Phoebus on more hardware than GitHub provides.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Ben Ryan ***@***.***>
Sent: Monday, December 4, 2023 8:23:16 AM
To: lanl/phoebus ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [lanl/phoebus] Long term solution for gold files (Issue #187)


[...]

PR #188, opened 2023-12-02
Fix & refector: Incorrect v^2 calc in some pgens, and move Reducers out of Driver class.

AstroBarker:
The name says it all. When I was trying to test tracers with the 2D advection problem, I eventually found that the `vsq` calculation was erroneous, showing the following pattern (psueudocode):

```
SPACELOOP2(ii, jj) { vsq += v(ii) * v(jj); }
```

Namely, it was either 1) missing the metric (which would collapse the sum in Minkowski or 2) should've been manually collapsed to just `SPACELOOP(ii)`. The result was `vsq > 1` and unphysical Lorentz factors. This PR fixes that by pulling in the metric (opting for the more general solution). I noticed the same pattern in the `kh` and `linear_modes` problems. `kh` is fixed the same, while `linear_modes` I replaced the loop with a 1D `SPACELOOP` to not fight with the existing metric for when using e.g., snake coordinates. 

There's a (somewhat) unrelated refactor in this PR. I was working on adding a Parthenon-style advection regression test #187 but ran into an issue: the reducer objects for e.g., `particle_resolution` in the Driver class would have their destructors called after MPI/Kokkos finalize and Phoebus would exit with an error code, which caused the regression testing to exit with an error, too. So, this PR also refactors those Reducer objects as mutable params into `radiation` and `fluid`, where appropriate. I also split off the Driver creation and execution into its own scope. Not necessary with the refactor, but safer and potentially more future-proof. In a separate PR I'll pull in the regression testing (see #187 ).

### DIFF
--- a/src/pgen/kh.cpp
+++ b/src/pgen/kh.cpp
@@ -66,6 +66,8 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
   auto emin = pmb->packages.Get("eos")->Param<Real>("sie_min");
   auto emax = pmb->packages.Get("eos")->Param<Real>("sie_max");
 
+  auto geom = Geometry::GetCoordinateSystem(rc.get());
+
   RNGPool rng_pool(pin->GetOrAddInteger("kelvin_helmholtz", "seed", 37));
 
   pmb->par_for(
@@ -77,6 +79,9 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
         const Real rho = y < 0.25 ? rho1 : rho0;
         const Real P = y < 0.25 ? P1 : P0;
         const Real vel = y < 0.25 ? v1 : v0;
+
+        Real gcov[4][4];
+        geom.SpacetimeMetric(0.0, x, y, 0.0, gcov);
 
         Real eos_lambda[2];
         if (iye > 0) {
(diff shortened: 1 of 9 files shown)


src/pgen/kh.cpp
```diff
@@ -78,6 +80,9 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
         const Real P = y < 0.25 ? P1 : P0;
         const Real vel = y < 0.25 ? v1 : v0;
 
+        Real gcov[4][4];
+        geom.SpacetimeMetric(0.0, x, y, 0.0, gcov);
```
Yurlungur, line 84: not that it matters much, but why set z to zero here? Why not use the z from the coords?

src/pgen/linear_modes.cpp
```diff
@@ -247,7 +247,7 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
         }
 
         Real vsq = 0.;
-        SPACELOOP2(ii, jj) { vsq += v(ivlo + ii, k, j, i) * v(ivlo + jj, k, j, i); }
+        SPACELOOP(ii) { vsq += v(ivlo + ii, k, j, i) * v(ivlo + ii, k, j, i); }
```
Yurlungur, line 250: is the metric not needed for this one?

Yurlungur (approved):
Thanks for the catches! LGTM

Yurlungur:
My comments are non-blockers, so I'm just going to merge. But may be wroth revisiting them at some point.

PR #190, opened 2023-12-06
Add tracer particles to phoebus

AstroBarker:
I believe that the tracer treatment is now mature enough to have more eyes on it. It's a pretty big PR and has been awhile coming, so interact in whatever level you feel.

## Summary
Here I add tracer particle support in phoebus. We create a tracer package that, if enabled, registers the appropriate swarm variables. See `src/tracers/tracers.cpp`. At present, the tracers are coupled in first order operator split, though in the future I'd like to explore improving this, after some phoebus task list refactoring. Tracers are advected in `AdvectTracers(...)`, interpolating the appropriate quantities to the tracer positions and then pushing (with RK2). The "filling" of the tracers is handled in `FillTracers` registered with `UserWorkBeforeOutput`. To enable this, I added `phoebus::UserWorkBeforeOutput` to the driver.
It's a lot of quantities.. and may desire to shorten later, or otherwise make what is stored runtime modular?

To deal with black hole accretion, when FMKS geometry is enabled, we use `PurgeParticles` in `src/fixup/fixup_particles.cpp` If FMKS is not enabled, the `constexpr if` resolves to a no op. In principle, here, future "purge particles if.." conditions can be supplied.

At present, I have implemented tracers in the `torus` and `advection` pgens. The advection test yields relative errors on tracer positions of ~ 1e-14 after one period. Tracer orbits in a non-magnetized torus work as desired.

To run with tracers (when the pgen supports it), simply 
```
<physics>
tracers = true

<tracers>
num_tracers = ...
```

## TODO:
- [x] Add tracers into advection, at least. Good to have a simpler-than-torus test problem.
- [x] Do 3D, non-magnetized torus orbit test.
- [ ] Unit & regression test.
image

cc: @bprather


src/pgen/torus.cpp
```diff
@@ -534,15 +547,115 @@ void PostInitializationModifier(ParameterInput *pin, Mesh *pmesh) {
...
+      auto &mass = swarm->Get<Real>("mass").Get();
+      auto &id = swarm->Get<int>("id").Get();
+
+      /* lambda to flatten (k, j, i) -> n */
+      auto flatten = KOKKOS_LAMBDA(const int k, const int j, const int i) {
+        return (i + (ib.e - ib.s + 1) * (j + (jb.e - jb.s + 1) * k));
+      };
+
+      // TODO (BLB): better way to construct & access indices array?
+      // Access as indices( flatten( k-kb.s, j-jb.s, i-ib.s ) )
```
Yurlungur, line 594: I like @brryan 's idea of using a "volume-weighted" mask, as discussed on MM. We should look at the tracers example parthenon.

Yurlungur: Note we don't have to fix this now. PR can go in even if the initialization is rough around the edges.

src/tracers/tracers.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  return TaskStatus::complete;
+} // AdvectTracers
+
+/**
+ * FillDerived function for tracers.
+ * Registered Quantities (in addition to t, x, y, z):
+ * rho, T, ye, vel, W_lorentz, pressure, lapse, shift, entropy, detgamma, B
+ **/
+void FillTracers(MeshBlockData<Real> *rc) {
```
Yurlungur, line 155: I was initially a little confused why `FillTracers` and `AdvectTracers` aren't a single kernel---but I think it makes sense to split things this way... In fact, in the long term, we should consider moving `FillTracers` from `FillDerivedBlock` to `UserWorkBeforeOutput`. That way this interpolation only happens on I/O, not every step. That's for future us, though. :)

Yurlungur (approved):
Overall this looks great! Thanks for your hard work @AstroBarker this is awesome. A few nitpicks below. And also tests seem to be not passing.

AstroBarker:
re: tests failing, I had some old testing code in `radiation_advection` and `radiation_equilibration` problems that didn't work with tracers disabled. I've just reverted those files to pre-tracers, and will put tracers in the fluid advection problem.

inputs/blast_wave_2d.pin
```diff
@@ -93,6 +93,7 @@ hydro = true
 he = false
 3t = false
 rad = false
+tracers = false
```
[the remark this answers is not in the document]
AstroBarker: Was using `GetBoolean` not `GetOrAddBoolean(_, _, false)`. Fixed!

src/pgen/torus.cpp
```diff
@@ -534,15 +547,115 @@ void PostInitializationModifier(ParameterInput *pin, Mesh *pmesh) {
...
+      const int mysize = (kb.e - kb.s + 1) * (jb.e - jb.s + 1) * (ib.e - ib.s + 1);
+      ParArray1D<int> indices("map from k,j,i to 1D particle array", mysize);
```
AstroBarker, line 596: Weirdness on GPU accessing the indices gymnastics @Yurlungur . In `Phoebus::ProblemGenerator::Torus::SumDiskCells` kernel, `n_cells_disk` is always 0.

Yurlungur: AH! I see the problem. `indices` is not guarded with an atomic. You need a `Kokkos::atomic` on the `indices` assignment. Alternatively, you need to make it a raw kokkos view with the `atomic` trait.

src/pgen/torus.cpp
```diff
@@ -534,15 +547,115 @@ void PostInitializationModifier(ParameterInput *pin, Mesh *pmesh) {
...
+            const Real x2 = coords.Xc<2>(k, j, i);
+            Real r = tr.bl_radius(x1);
+            Real th = tr.bl_theta(x1, x2);
+
+            Real lnh = -1.0;
+            Real hm1;
+            Real uphi;
+            if (r > rin) lnh = log_enthalpy(r, th, a, rin, angular_mom, uphi);
+            if (lnh > 0.0 && x1 > xh) {
+              indices(flatten(k - kb.s, j - jb.s, i - ib.s)) = n_cells_disk;
```
Yurlungur, line 613: `n_cells_disk` is not guaranteed to be assigned  in any particular order... so making `indices` have atomic writes may not solve this problem. But let's try that first.

AstroBarker:
Some updates:
 - I have refactored the torus pgen to use rejection sampling. It is cleaner now, and doesn't fail on GPU. As implemented, it should be uniform in volume.
 - I have split the tracer rhs calculation into a separate function
 - Tracer integration is now RK2, as in nubhlight
 - Added Bernoulli parameter `- u^0 h - 1` to tracer output.
 - Tracers are implemented into the advection pgen
   - Implementation here could be improved. As is, we get nblocks * num_tracers particles. Should just move to post init...
 
In the advection test, relative errors on tracer positions was ~ 1e-14 or lower after one period.
I also ran a 3D, non-magnetized torus to see tracer orbits (notes: below only goes to t=500.. longer is running. And this was before the upgrade to RK2)

tr_eq

Yurlungur:
Looks awesome! When do you think you'll be ready to remove the WIP markings? :)

AstroBarker:
One last change: I've moved the `FillTracers` call from `PostFillDerivedBlock` into `UserWorkBeforeOutput`. To accommodate this, I created `phoebus::UserWorkBeforeOutput` in the driver and registered it in main. At present it just unpacks `MeshBlockData` and calls `FillTracers`, but it could accommodate extension in the future. Should cut down on unnecessary work. Let me know how this bit looks @Yurlungur.

(Also fixed something I forgot to commit in the torus init..).

src/tracers/tracers.cpp
Yurlungur: This is now done, right? :)

src/phoebus_driver.cpp
```diff
@@ -991,4 +1027,17 @@ TaskListStatus PhoebusDriver::MonteCarloStep() {
...
+ * Currently: Fills tracers.
+ **/
+void UserWorkBeforeOutput(MeshBlock *pmb, ParameterInput *pin) {
+  auto tracer_pkg = pmb->packages.Get("tracers");
+  bool do_tracers = tracer_pkg->Param<bool>("active");
+  if (do_tracers) {
+    auto &mbd = pmb->meshblock_data.Get();
+    tracers::FillTracers(mbd.get());
+  }
+}
```
Yurlungur, line 1041: :+1: When particle meshblock packs work, we should refactor to use those :)

src/pgen/torus.cpp
```diff
@@ -532,18 +545,178 @@ void PostInitializationModifier(ParameterInput *pin, Mesh *pmesh) {
...
+            const Real x1 = coords.Xc<1>(k, j, i);
+            const Real x2 = coords.Xc<2>(k, j, i);
+            Real r = tr.bl_radius(x1);
+            Real th = tr.bl_theta(x1, x2);
+
+            Real lnh = -1.0;
+            Real uphi;
+            if (r > rin) lnh = log_enthalpy(r, th, a, rin, angular_mom, uphi);
+            if (lnh > 0.0) {
+              const Real vol = coords.CellVolume(k, j, i);
```
Yurlungur, line 588: This should have a factor of the metric in it I think. (It probably doesn't matter for problems we run, but we should include it.)

src/pgen/torus.cpp
```diff
@@ -532,18 +545,178 @@ void PostInitializationModifier(ParameterInput *pin, Mesh *pmesh) {
...
+            const Real x2 = coords.Xc<2>(k, j, i);
+            Real r = tr.bl_radius(x1);
+            Real th = tr.bl_theta(x1, x2);
+
+            Real lnh = -1.0;
+            Real uphi;
+            if (r > rin) lnh = log_enthalpy(r, th, a, rin, angular_mom, uphi);
+            if (lnh > 0.0 && x1 > xh) {
+              const Real vol = coords.CellVolume(k, j, i);
+              number_block_reduce += vol;
```
Yurlungur, line 659: factor of metric determinant missing (see above)

src/pgen/torus.cpp
```diff
@@ -532,18 +545,178 @@ void PostInitializationModifier(ParameterInput *pin, Mesh *pmesh) {
...
+
+      // Now calculate the mass per tracer as mass(cell) / num_tracers_cell
+      // Separate kernel, as it requires knowing the particle count per cell
+      pmb->par_for(
+          "ProblemGenerator::Torus::DistributeTracers", 0, max_active_index,
+          KOKKOS_LAMBDA(const int n) {
+            if (swarm_d.IsActive(n)) {
+              int k, j, i;
+              swarm_d.Xtoijk(x(n), y(n), z(n), i, j, k);
+              const Real vol = coords.CellVolume(k, j, i);
```
Yurlungur, line 714: missing detgamma

src/pgen/advection.cpp
```diff
@@ -102,6 +104,59 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
+
+            // sample in ye ball
+            Real r2 = 1.0 + rin * rin; // init > rin^2
+            while (r2 > rin * rin) {
+              x(n) = x_min + rng_gen.drand() * (x_max - x_min);
+              y(n) = y_min + rng_gen.drand() * (y_max - y_min);
+              z(n) = z_min + rng_gen.drand() * (z_max - z_min);
+              r2 = x(n) * x(n) + y(n) * y(n) + z(n) * z(n);
+            }
+            id(n) = num_tracers_total * gid + n;
```
Yurlungur, line 152: :+1:

Yurlungur (approved):
This looks great, @AstroBarker ! I'm happy to merge this, but I would like @brryan 's eyes on it if he gets the chance to look.

Also this is a question for Ben: Are particle IDs integers? Or longs? We might want to make them longs if possible (which might be a parthenon change). I recall hitting integer overflow in nubhlight.

Yurlungur:
Actually @AstroBarker before merge, please correct the missing metric factor in the volumes in the torus calculation.

Yurlungur:
Since @brryan never had a chance to look. I'm going to click the button. Ben whenever you get the chance, I'd still like your thoughts.

PR #203, opened 2024-03-07
central entropy output added

mari2895:
## PR Summary
added output for entropy at z=0

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/analysis/analysis.cpp
+++ b/src/analysis/analysis.cpp
@@ -6,6 +6,9 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   auto analysis_pkg = std::make_shared<StateDescriptor>("analysis");
   Params &params = analysis_pkg->AllParams();
   Real sigma = pin->GetOrAddReal("analysis", "sigma", 2e-3);
+  if (sigma < 0){
+    PARTHENON_THROW("sigma must be greater than 0");
+  }
   params.Add("sigma", sigma);
 
   return analysis_pkg;
--- a/src/analysis/analysis.cpp
+++ b/src/analysis/analysis.cpp
@@ -1,5 +1,5 @@
-#include "phoebus_utils/variables.hpp"
 #include "analysis.hpp"
+#include "phoebus_utils/variables.hpp"
 
 namespace analysis{
 std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
@@ -9,5 +9,5 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   params.Add("sigma", sigma);
 
   return analysis_pkg;
-} //initialize
-} //namespace analysis
+} // initialize
+} // namespace analysis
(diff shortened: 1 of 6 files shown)


src/phoebus_driver.cpp
```diff
@@ -1068,6 +1069,10 @@ void UserWorkBeforeOutput(MeshBlock *pmb, ParameterInput *pin) {
         Real gam1[3][3];
         Real gam2[3][3];
         Real gam3[3][3];
+        const Real z = coords.Xc<3>(k, j, i);
+        const Real sigma = 2 * coords.CellWidthFA(X3DIR, k, j, i);
```
Yurlungur, line 1073: this depends on the mesh width on the block you're on which makes me a bit uncomfortable. I suggest making sigma a free parameter that is set-able in the input deck.

src/phoebus_driver.cpp
```diff
@@ -1068,6 +1069,10 @@ void UserWorkBeforeOutput(MeshBlock *pmb, ParameterInput *pin) {
...
+        const Real s0 = s * std::exp(-z * z / sigma / sigma);
```
Yurlungur, line 1074: Alternatively (or possibly additionally) we should probably normalize our Gaussian

Yurlungur (changes requested):

src/phoebus_driver.cpp
mari2895: hmm that's a weird thing to have in input deck... plus out of context from everything else, no?

src/phoebus_driver.cpp
mari2895: done

src/phoebus_driver.cpp
mari2895: done

src/fluid/fluid.cpp
```diff
@@ -207,6 +207,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   physics->AddField(p::entropy::name(), mprim_scalar);
   physics->AddField(p::cs::name(), mprim_scalar);
   physics->AddField(diag::ratio_divv_cs::name(), mprim_scalar);
+  physics->AddField(diag::entropy_z_0::name(), mprim_scalar);
```
Yurlungur, line 210: might make more sense to register this var in the analysis initialize function. But this is fine for now.

src/analysis/analysis.cpp
```diff
@@ -0,0 +1,13 @@
+#include "phoebus_utils/variables.hpp"
+#include "analysis.hpp"
+
+namespace analysis{
+std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
+  auto analysis_pkg = std::make_shared<StateDescriptor>("analysis");
+  Params &params = analysis_pkg->AllParams();
+  Real sigma = pin->GetOrAddReal("analysis", "sigma", 2e-3);
+  params.Add("sigma", sigma);
```
Yurlungur, line 9: Maybe add a check that sigma > 0

Yurlungur (approved):

src/analysis/analysis.cpp
mari2895: done

PR #205, opened 2024-03-11
outputs for SN postprocessing: max density, M_gain, and Qdot_gain

mari2895:
## PR Summary
Created a reducer that only integrates in the gain region. Added the following outputs in history file: max density (central density, mass_gain and total net neutrino heating in the gain region). 

## PR Checklist

- [x] Explain what you did.


src/analysis/history.cpp
```diff
@@ -235,4 +236,46 @@ Real ReduceMagneticFluxPhi(MeshData<Real> *md) {
   return 0.5 * result; // 0.5 \int detg B^r dx2 dx3
 } // Phi
 
+// SN analysis
+
+void ReduceLocalizationFunction(MeshData<Real> *md) {
```
Yurlungur, line 241: Maybe add a comment above this indicating it's not currently in use?

src/analysis/history.hpp
```diff
@@ -58,21 +61,71 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
   // This assumes that for "Sum" variables, the variable is densitized, meaning
   // it already contains a factor of the measure: sqrt(det(gamma))
+  Real result = 0.0;
+  Reducer_t reducer(result);
+
   const bool volume_weighting =
-      std::is_same<Reducer_t, Kokkos::Sum<Real, HostExecSpace>>::value;
+      std::is_same<Reducer_t, Kokkos::Sum<Real, HostExecSpace>>::value ||
+      std::is_same<Reducer_t, Kokkos::Sum<Real, DevExecSpace>>::value ||
+      std::is_same<Reducer_t, Kokkos::Sum<Real, Kokkos::DefaultExecutionSpace>>::value;
```
Yurlungur, line 70: :+1:

src/analysis/history.hpp
```diff
@@ -58,21 +61,71 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+        const auto &coords = pack.GetCoords(b);
+        const Real vol = volume_weighting ? coords.CellVolume(k, j, i) : 1.0;
+        reducer.join(lresult, pack(b, ivar.first + idx, k, j, i) * vol);
+      },
+      reducer);
+  return result;
+}
+
+template <typename Reducer_t>
+Real ReduceInGain(MeshData<Real> *md, const std::string &varname, int idx = 0) {
```
Yurlungur, line 88: I don't think `ReduceInGain` needs to be templated on reducer type, since we're always going to be summing. It probably also doesn't need to take in `varname` or an `index`. I would make it just take `MeshData`. If it's not templated, you can also move the implementation to `history.cpp` instead of `hpp` and remove some of the includes above.

src/analysis/history.hpp
```diff
@@ -58,21 +61,71 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+  const auto kb = md->GetBoundsK(IndexDomain::interior);
+
+  PackIndexMap imap;
+  namespace iv = internal_variables;
+  std::vector<std::string> vars = {varname, iv::GcovHeat::name(), iv::GcovCool::name()};
+  const auto pack = md->PackVariables(vars, imap);
+  const auto ivar = imap[varname];
+  const auto iheat = imap[iv::GcovHeat::name()].first;
+  const auto icool = imap[iv::GcovCool::name()].first;
+  auto *pmb = md->GetParentPointer();
```
Yurlungur, line 100: switch to sparse packs. Also just pull both heat and cool out here, rather than passing them in is as string arguments above.

src/analysis/history.hpp
```diff
@@ -58,21 +61,71 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+  PARTHENON_REQUIRE(ivar.first >= 0, "Var must exist");
+  PARTHENON_REQUIRE(ivar.second >= ivar.first + idx, "Var must exist");
```
Yurlungur, line 102: not needed if you follow above advice

src/analysis/history.hpp
```diff
@@ -58,21 +61,71 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+  // We choose to apply volume weighting when using the sum reduction.
+  // This assumes that for "Sum" variables, the variable is densitized, meaning
+  // it already contains a factor of the measure: sqrt(det(gamma))
   Real result = 0.0;
   Reducer_t reducer(result);
+
+  const bool volume_weighting =
+      std::is_same<Reducer_t, Kokkos::Sum<Real, HostExecSpace>>::value ||
+      std::is_same<Reducer_t, Kokkos::Sum<Real, DevExecSpace>>::value ||
+      std::is_same<Reducer_t, Kokkos::Sum<Real, Kokkos::DefaultExecutionSpace>>::value;
```
Yurlungur, line 113: in this case, jsut always volume weight as reduction over gain will always be a sum

src/analysis/history.hpp
```diff
@@ -58,21 +61,71 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
         // reduction operation locally
+        Real gdet = geom.DetGamma(CellLocation::Cent, 0, k, j, i);
+        bool is_netheat = (pack(b, iheat, k, j, i) - pack(b, icool, k, j, i) > 0);
         const auto &coords = pack.GetCoords(b);
         const Real vol = volume_weighting ? coords.CellVolume(k, j, i) : 1.0;
-        reducer.join(lresult, pack(b, ivar.first + idx, k, j, i) * vol);
+        reducer.join(lresult,
+                     pack(b, ivar.first + idx, k, j, i) * gdet * vol * is_netheat);
       },
       reducer);
```
Yurlungur, line 130:
```suggestion
     Kokkos::Sum<Real>(result));
```

src/analysis/history.hpp
```diff
@@ -58,21 +61,71 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
         // join is a Kokkos construct
         // that automatically does the
         // reduction operation locally
+        Real gdet = geom.DetGamma(CellLocation::Cent, 0, k, j, i);
+        bool is_netheat = (pack(b, iheat, k, j, i) - pack(b, icool, k, j, i) > 0);
         const auto &coords = pack.GetCoords(b);
         const Real vol = volume_weighting ? coords.CellVolume(k, j, i) : 1.0;
-        reducer.join(lresult, pack(b, ivar.first + idx, k, j, i) * vol);
+        reducer.join(lresult,
+                     pack(b, ivar.first + idx, k, j, i) * gdet * vol * is_netheat);
```
Yurlungur, line 128: If you switch to sparse packs, remove the single var string argument above, and always assume summation, rather than arbitrary reduciton, this can be simplified to:
```C++
       lresult += gdet * vol * is*netheat * (pack(b, heat(), k, j, i) - pack(b, cool(), k, j, i));
```

src/analysis/history.hpp
```diff
@@ -17,12 +17,14 @@
 #include <string>
 #include <vector>
 
+#include "geometry/geometry.hpp"
+#include "geometry/geometry_utils.hpp"
+#include "phoebus_utils/variables.hpp"
 #include <kokkos_abstraction.hpp>
+#include <parthenon/driver.hpp>
```
Yurlungur, line 24: Some of these new includes are causing one of the unit tests to fail (for a stupid technical reason). Two ways to avoid: 
1. fix the build of the unit test
2. move your new reduction implementation to the `history.cpp` file and remove these includes.

I suggest 2 is easier.

src/fluid/fluid.cpp
```diff
@@ -319,6 +322,23 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   auto ReduceEn = [](MeshData<Real> *md) {
     return ReduceOneVar<Kokkos::Sum<Real>>(md, fluid_cons::energy::name(), 0);
   };
+  auto MaxDensitySN = [](MeshData<Real> *md) {
+    return ReduceOneVar<Kokkos::Max<Real>>(md, fluid_prim::density::name(), 0);
+  };
+  auto Mgain = [](MeshData<Real> *md) {
+    return ReduceInGain<Kokkos::Sum<Real, HostExecSpace>>(md, fluid_prim::density::name(),
+                                                          0);
+  };
```
Yurlungur, line 331: :+1:

src/fluid/fluid.cpp
```diff
@@ -319,6 +322,23 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  auto Qgain = [](MeshData<Real> *md) {
+    return ReduceInGain<Kokkos::Sum<Real, HostExecSpace>>(
+               md, internal_variables::GcovHeat::name(), 0) -
+           ReduceInGain<Kokkos::Sum<Real, HostExecSpace>>(
+               md, internal_variables::GcovCool::name(), 0);
+  };
```
Yurlungur, line 337: See my above comments.

src/fluid/fluid.cpp
```diff
@@ -319,6 +322,23 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+
+  hst_vars.emplace_back(HistoryOutputVar(HstMax, MaxDensitySN, "maximum density SN"));
```
Yurlungur, line 339: This is generic so it's maybe worth always outputting. You should make this one say "maximum density" it's not just for supernovae.

src/fluid/fluid.cpp
```diff
@@ -319,6 +322,23 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  hst_vars.emplace_back(HistoryOutputVar(HstSum, Mgain, "Mgain"));
+  hst_vars.emplace_back(HistoryOutputVar(HstSum, Qgain, "total net heat"));
```
Yurlungur, line 341: These should not be on by default, as they'll interfere with, e.g.,  simulations by @AstroBarker and @brryan . A few possible solutions:
1. Add a flag in `fluid.cpp` that turns them on or off depedning on the input deck
2. Move them into the progenitor package so they won't be active for non-SN sims

I suggest (2) but (1) is fine if you prefer.

Yurlungur (changes requested):
@mari2895 this looks good, but I think it needs some revision. Please let me know if you're confused by any of the blow comments.

src/analysis/history.hpp
mari2895: Why should not it take varname or index? Should not it be a general function that integrates over the gain region whatever variable we pass? sometimes we'll integrate scalar variable sometimes vector no?

src/analysis/history.hpp
mari2895: Yeah I know, it is the `geometry.hpp` flag. Why though?

src/analysis/history.hpp
Yurlungur: Ah I see--there are multiple things you may want to reduce over the gain region? Then yes that's fine. You can keep the string machinery + index machinery as is.

src/analysis/history.hpp
Yurlungur: `compile_constants.hpp` is generated automatically by the cmake. The unit tests don't automatically use or rely on it. This could be fixed but would require some build system work.

src/analysis/history.hpp
mari2895: Yes correct, okay.

src/analysis/history.cpp
mari2895: done

src/analysis/history.hpp
mari2895: done

src/analysis/history.hpp
mari2895: done

src/analysis/history.hpp
mari2895: done

src/analysis/history.hpp
mari2895: done

src/analysis/history.hpp
```diff
@@ -78,6 +84,42 @@ Real ReduceOneVar(MeshData<Real> *md, const std::string &varname, int idx = 0) {
...
+
+  namespace iv = internal_variables;
+  using parthenon::MakePackDescriptor;
+  Mesh *pmesh = md->GetMeshPointer();
+  auto &resolved_pkgs = pmesh->resolved_packages;
+  const int ndim = pmesh->ndim;
+  static auto desc =
+      MakePackDescriptor<Varname, iv::GcovHeat, iv::GcovCool>(resolved_pkgs.get());
+  auto v = desc.GetPack(md);
+  const int nblocks = v.GetNBlocks();
```
Yurlungur, line 101: Should probably have a check that heat and cool are present. Something like...
```C++
PARTHENON_REQUIRE_THROWS(v.ContainsHost(0, iv::GcovHeat(), iv::GcovCool()), "Must be doing SN simulation");
```

src/fluid/fluid.cpp
```diff
@@ -207,6 +208,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   physics->AddField(p::entropy::name(), mprim_scalar);
   physics->AddField(p::cs::name(), mprim_scalar);
   physics->AddField(diag::ratio_divv_cs::name(), mprim_scalar);
+  physics->AddField(diag::localization_function::name(), mprim_scalar);
```
Yurlungur, line 211: we're not using the localization function right? Let's comment this out for now and only add it when we need it.

Yurlungur (approved):
Looks good. Just two more nitpicks...

src/analysis/history.hpp
mari2895: done

Yurlungur (approved):

PR #210, opened 2024-03-19
Add eos information to floors, misc UserWorkBeforeOutput fix

AstroBarker:
This PR adds knowledge of eos bounds to the floors. Specifically, for tabulated EOS, demand that the floors do not extrapolate off of the table. This required making the `bounds` param mutable.

Currently, for ideal gas, the option to enforce user set bounds is supported by supplying, e.g., `<eos>/rho_min` in the input deck. If unspecified, reasonable defaults are set.

Also fixed a misc bug in the `UserWorkBeforeOutput` code where a package and param were pulled out inside a device kernel that caused a crash on device.

### DIFF
--- a/src/fixup/fixup.hpp
+++ b/src/fixup/fixup.hpp
@@ -208,9 +208,8 @@ class Ceilings {
   }
 
  private:
-  Real g0_, s0_, sie_max_eos_;
+  Real g0_, s0_;
   const int ceiling_flag_;
-  bool eos_bnds_set_;
 };
 
 static struct ConstantBsqRatCeiling {
--- a/src/fixup/fixup.hpp
+++ b/src/fixup/fixup.hpp
@@ -290,7 +290,7 @@ class Bounds {
         radiation_floors_(RadiationFloors()), radiation_ceilings_(RadiationCeilings()) {}
 
   template <class... Args>
-  KOKKOS_INLINE_FUNCTION void SetEOSBnds(Args &&...args) {
+  void SetEOSBnds(Args &&...args) {
     floors_.SetEOSBnds(std::forward<Args>(args)...);
   }
 
--- a/src/fixup/fixup.hpp
+++ b/src/fixup/fixup.hpp
@@ -91,37 +91,60 @@ class Floors {
   KOKKOS_INLINE_FUNCTION
   void GetFloors(const Real x1, const Real x2, const Real x3, Real &rflr,
                  Real &sflr) const {
+    if (!eos_bnds_set_) {
+      PARTHENON_FAIL("EOS bounds not set in floors.");
+    }
+
     switch (floor_flag_) {
     case FloorFlag::ConstantRhoSie:
       rflr = r0_;
       sflr = s0_;
+      rflr = std::max(rflr, rho_min_eos_);
+      sflr = std::max(sflr, sie_min_eos_);
       break;
     case FloorFlag::ExpX1RhoSie:
       rflr = r0_ * std::exp(ralpha_ * x1);
       sflr = s0_ * std::exp(salpha_ * x1);
+      rflr = std::max(rflr, rho_min_eos_);
+      sflr = std::max(sflr, sie_min_eos_);
       break;
     case FloorFlag::ExpX1RhoU: {
       Real scratch = r0_ * std::exp(ralpha_ * x1);
       sflr = s0_ * std::exp(salpha_ * x1) / std::max(rflr, scratch);
       rflr = scratch;
+      rflr = std::max(rflr, rho_min_eos_);
+      sflr = std::max(sflr, sie_min_eos_);
     } break;
     case FloorFlag::X1RhoSie:
       rflr = r0_ * std::min(1.0, std::pow(x1, ralpha_));
       sflr = s0_ * std::min(1.0, std::pow(x1, salpha_));
+      rflr = std::max(rflr, rho_min_eos_);
+      sflr = std::max(sflr, sie_min_eos_);
       break;
     case FloorFlag::RRhoSie: {
       Real r = std::sqrt(x1 * x1 + x2 * x2 + x3 * x3);
       rflr = r0_ * std::min(1.0, std::pow(r, ralpha_));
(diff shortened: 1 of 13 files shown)


src/fluid/fluid.cpp
```diff
@@ -324,11 +319,6 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   auto ReduceEn = [](MeshData<Real> *md) {
     return ReduceOneVar<Kokkos::Sum<Real>>(md, fluid_cons::energy::name(), 0);
   };
-  auto MaxDensity = [](MeshData<Real> *md) {
-    return ReduceOneVar<Kokkos::Max<Real>>(md, fluid_prim::density::name(), 0);
-  };
-
-  hst_vars.emplace_back(HistoryOutputVar(HstMax, MaxDensity, "maximum density"));
```
Yurlungur, line 331: Looks like this branch might be out of date with main... double check you and @mari2895 are in sync so your PRs don't undo each other.

Yurlungur (changes requested):
I think i see why you're getting segfaults. See comments below. After making those changese everywhere relevant (I only highlighted once), give it a try on device again.

src/fixup/fixup.hpp
```diff
@@ -91,37 +91,60 @@ class Floors {
   KOKKOS_INLINE_FUNCTION
   void GetFloors(const Real x1, const Real x2, const Real x3, Real &rflr,
                  Real &sflr) const {
+    if (!eos_bnds_set_) {
+      PARTHENON_FAIL("EOS bounds not set in floors.");
+    }
```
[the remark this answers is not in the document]
AstroBarker: At present it does -- the ideal gas sets the same `rho_min` variables etc. At present, they default to zero and [some large number] for the min and max. However, they can be specified in the `<eos>` block of the input deck to some other value. I am not sure if we actually want that behavior, or to just set the bounds in this case to small/large values. Thoughts?

src/fixup/fixup.hpp
```diff
@@ -173,19 +196,30 @@ class Ceilings {
   KOKKOS_INLINE_FUNCTION
   void GetCeilings(const Real x1, const Real x2, const Real x3, Real &gmax,
                    Real &smax) const {
+    if (!eos_bnds_set_) {
+      PARTHENON_FAIL("EOS bounds not set in ceilings.");
+    }
```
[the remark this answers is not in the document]
AstroBarker: Fair enough. I was thinking of density when I thought of the ceiling, which gets quite weird at the top of the table, but that isn't enforced in the ceilings anyway. Probably safe to remove this then.

src/fixup/fixup.hpp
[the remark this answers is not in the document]
Yurlungur: I think it's fine to inherit from the bounds that the eos sets, so long as it sets sensible things.

Yurlungur:
Ready for re-review, @AstroBarker ?

AstroBarker:
> Ready for re-review, @AstroBarker ?

Good to go

src/fluid/fluid.cpp
AstroBarker: Looks like this is the same as current main, unless I missed something.

src/fixup/fixup.hpp
```diff
@@ -184,8 +208,9 @@ class Ceilings {
   }
 
  private:
-  Real g0_, s0_;
+  Real g0_, s0_, sie_max_eos_;
```
Yurlungur, line 211: `sie_max_eos_` no longer needed right?

src/fixup/fixup.hpp
```diff
@@ -184,8 +208,9 @@ class Ceilings {
...
   const int ceiling_flag_;
+  bool eos_bnds_set_;
```
Yurlungur, line 213: no longer needed?

src/fluid/fluid.cpp
Yurlungur: Why is there a diff showing here then? Odd.

Yurlungur (approved):
Two tiny nitpicks. Otherwise this is good to go.

src/fluid/fluid.cpp
AstroBarker: Oh I see it now... not sure why it's different from main, don't think I did this myself..

Yurlungur: maybe just merge main in again... see if that fixes it?

mari2895: @AstroBarker Looks the same on my branch too. Don't see anything different.

mari2895: but why is it deleted? should not be deleted

src/phoebus_driver.cpp
```diff
@@ -1071,9 +1073,7 @@ void UserWorkBeforeOutput(MeshBlock *pmb, ParameterInput *pin) {
         Real gam1[3][3];
         Real gam2[3][3];
         Real gam3[3][3];
-        auto analysis = pmb->packages.Get("analysis").get();
         const Real z = coords.Xc<3>(k, j, i);
-        const Real sigma = analysis->Param<Real>("sigma");
```
mari2895, line 1076: why are you deleting these lines? I need sigma

AstroBarker: I moved them out of the `par_for` as this code crashes on device. Packages and params are only allocated on the CPU.

mari2895: oh shoot, I did too in my PR.

mari2895: But I don't think that's a problem, maybe one of us can undo this change and whichever PR is merged first merge that main into another PR. @Yurlungur what do you think?

Yurlungur: There's no need to undo changes. Whoever's PR gets merged first, git will guide you through a manual merge process if there are merge conflicts on the second PR.

mari2895: Okay then even better

Yurlungur (approved):

mari2895 (approved):
LGTM, thanks for explaining!

PR #212, opened 2024-03-25
Trivial: update singularity-opac to latest

AstroBarker:
[Recently](https://github.com/lanl/singularity-opac/pull/50) I updated `singularity-opac` to fix the submodules to pull `ports-of-call` via https instead of ssh. This PR pins `singularity-opac` to latest so that we don't hit any ssh when cloning


mari2895 (approved):
Thanks!

carlnotsagan:
thanks for taking care of this!

Yurlungur (approved):

PR #218, opened 2024-10-14
Run on Frontier: update singularity, relocatable device code, misc fixes

AstroBarker:
This PR incorporates changes I needed to make to run on Frontier, plus some misc fixes:

- Update `singularity-eos` to main
- Move `energy_from_rho_P` from `pgen.cpp` into the header file.
- The `mckinnie_gammie_ryan` constructor had a bug in a variable name
- `eos_phoebus.cpp` had an incorrect redefinition of the `lambda` array that the compiler interpreted as an out of bounds array access
- In the sedov pgen, the boolean to use spherical coords was loaded from input with `GetOrAddReal`

### DIFF
--- a/src/pgen/pgen.hpp
+++ b/src/pgen/pgen.hpp
@@ -138,8 +138,8 @@ class PressResidual {
 
 template <typename T>
 KOKKOS_INLINE_FUNCTION Real energy_from_rho_P(T &eos, const Real rho, const Real P,
-                                       const Real emin, const Real emax,
-                                       const Real Ye = 0.0) {
+                                              const Real emin, const Real emax,
+                                              const Real Ye = 0.0) {
   PARTHENON_REQUIRE(P >= 0, "Pressure is negative!");
   PressResidual res(eos, rho, P, Ye);
   root_find::RootFind root;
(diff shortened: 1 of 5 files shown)


src/pgen/pgen.hpp
```diff
@@ -117,9 +118,34 @@ static std::map<std::string, std::function<void(ParameterInput *pin, Mesh *pmesh
...
+  }
+
+ private:
+  const EOS &eos_;
+  Real rho_, P_;
+  Real lambda_[2];
+};
+
+template <typename T>
+KOKKOS_FUNCTION Real energy_from_rho_P(T &eos, const Real rho, const Real P,
```
Yurlungur, line 140: `KOKKOS_INLINE_FUNCTION` now.

Yurlungur (approved):
Thanks for the fixes! How bad are compile times?

AstroBarker:
> Thanks for the fixes! How bad are compile times?

Not speedy but not miserable -- I've seen much worse on chicoma

Yurlungur:
Also looks like format tests are failing :)

PR #228, opened 2025-02-12
Add a tracer advection problem to regression suite

AstroBarker:
Small PR to add a tracer advection test. Runs the 2D advection problem with tracers using a fixed random seed. I also add type hinting to the regressing testing script. I think it improves readability.


tst/regression/regression_test.py
```diff
@@ -224,32 +231,68 @@ def cleanup():
...
+        Run with MPI (default is False).
+    build_type : str, optional
+        CMake build type (default is "Release").
+    upgold : bool, optional
+        Update gold file (default is False).
+    compression_factor : int, optional
+        Compression factor for gold file storage  (default is 1).
+    tolerance : float, optional
+        Error tolerance (default is 1e-5).
+    """
```
Yurlungur, line 295: ooh nice

Yurlungur (approved):
Great to have this in.

PR #229, opened 2025-02-12
Move tracer send/receives to MeshData level

AstroBarker:
This moves the tracer particles communication routines (`ResetCommunication`, `Send`, and `Receive`) to be `MeshData` operations, and implements a swarm `Defrag` on `MeshData`.

PR #230, opened 2025-02-19
Swarm regression test fix, device bug fix, misc cleanup

AstroBarker:
Seems that potentially #229 changed some behavior of tracers, not preserving ordering on multiple blocks. Nothing major, but caused a test failure due to not comparing positions of identical tracers. This edits the regression testing to sort tracers by our 'id', ensuring that the same tracers are compared against one another.

It also fixed a couple of errors that crept in through #225 :
- Erroneously made a variable needed on device 'static'. Doesn't work.
- Left a couple of testing tensor swarm variables in tracers that should have been deleted prior to committing..... don't look


Yurlungur (approved):
Thanks for the fix!

PR #233, opened 2025-03-04
Fixes to the monopole solver

Yurlungur:
## PR Summary

As suggested by @mari2895 . Thanks, Mariam!

## PR Checklist

- [ ] Explain what you did.

### DIFF
--- a/src/fluid/con2prim_robust.hpp
+++ b/src/fluid/con2prim_robust.hpp
@@ -62,7 +62,7 @@ class Residual {
   }
 
   KOKKOS_FORCEINLINE_FUNCTION
-  Real x_mu(const Real mu) { return 1.0 / (1.0 + mu * bsq_); }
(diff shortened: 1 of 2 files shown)


AstroBarker:
This LGTM, looks like it needs some formatting

curiousmiah:
Based upon my recollection (which is weak), these GR fixes seem about right to me.  I'd double check with Mariam.

Yurlungur:
I pulled these from Mariam's branch, so they should be what she implemented. That said, a double check may be worthwhile.

mari2895 (approved):
This matches what I have in my notes as well.

mari2895:
I can see i have more changes that i have not committed, but the only reason I would not commit those would be that I was not sure in the correctness of those fixes. So it is safe to have this only merge first.

AstroBarker (commented):
I think this looks good to me. We can iterate on any additional fixes @mari2895 might have found.

PR #236, opened 2025-04-24
Added specific internal energy in the output in primitive quantities.

mari2895:
## PR Summary
Added specific internal energy in the output in primitive quantities.

## PR Checklist

- [ ] Explain what you did.

### DIFF
--- a/src/fluid/fluid.cpp
+++ b/src/fluid/fluid.cpp
@@ -210,6 +210,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   physics->AddField(p::pressure::name(), mprim_scalar);
   physics->AddField(p::temperature::name(), mprim_scalar);
   physics->AddField(p::entropy::name(), mprim_scalar);
+  physics->AddField(p::sie::name(), mprim_scalar);
   physics->AddField(p::cs::name(), mprim_scalar);
   physics->AddField(diag::ratio_divv_cs::name(), mprim_scalar);
   // physics->AddField(diag::localization_function::name(), mprim_scalar);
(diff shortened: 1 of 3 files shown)


Yurlungur (approved):
I don't object to this, though I don't think it's really necessary. sie is just u/rho, after all. That said, this is more convenient.

mari2895:
> I don't object to this, though I don't think it's really necessary. sie is just u/rho, after all. That said, this is more convenient.

Do we output u? We have c.energy but I assumed that was a total conserved energy.

mari2895:
> `p.energy` is `u`

oh true we have p.energy. Is it without kinetic energy? Well even if it with kinetic energy we can just subtract I guess. Then maybe we can ignore this PR? The reason I added was that it was suggested during our sync meeting that to add cause it would be useful diagnostic quantity but either way (outputting or calculating) is fine for me.

Yurlungur:
> > `p.energy` is `u`
> 
> oh true we have p.energy. Is it without kinetic energy? Well even if it with kinetic energy we can just subtract I guess. Then maybe we can ignore this PR? The reason I added was that it was suggested during our sync meeting that to add cause it would be useful diagnostic quantity but either way (outputting or calculating) is fine for me.

`p.energy` is internal energy by volume, not kinetic. That said, there's no reason not to include the specific internal energy in the output for convenience. I will merge this.

mari2895:
> > > `p.energy` is `u`
> > 
> > 
> > oh true we have p.energy. Is it without kinetic energy? Well even if it with kinetic energy we can just subtract I guess. Then maybe we can ignore this PR? The reason I added was that it was suggested during our sync meeting that to add cause it would be useful diagnostic quantity but either way (outputting or calculating) is fine for me.
> 
> `p.energy` is internal energy by volume, not kinetic. That said, there's no reason not to include the specific internal energy in the output for convenience. I will merge this.

Good to know, thanks Jonah!

PR #240, opened 2025-08-22
New diagnostic capabilities in phoebuspy.py and Phoebus

curiousmiah:
## PR Summary

This PR provides new diagnostic tools in phoebuspy.py.  In particular it includes capabilities to plot and make movies of slices of 3D simulations.  It includes 2D slices and 1D slices.  In addition, there was a bug in Phoebus when including flux_divergence and src_terms in the output.  This has been fixed.

## PR Checklist

- [ ] Explain what you did.

### DIFF
--- a/scripts/python/phoebuspy.py
+++ b/scripts/python/phoebuspy.py
@@ -132,7 +132,7 @@ def ReadParameterFile(FileName='Params.dat'):
             params[fpar] = float(params[fpar])
 
     #convert to int
-    intparams = ['sliceaxis']
+    intparams = ['sliceaxis','nslices']
     for ipar in intparams:
         if(ipar in params):
             params[ipar] = int(params[ipar])
@@ -387,6 +387,52 @@ def Make2DSlice(data,sliceaxis=1,slice=0.,extractvars=['p.density']):
 
     return slice_data
 
+def MakeSlicesVsTime(params):
+    import pickle
+    #List of outfile names                                                      
+    filenames = sorted(glob(f"*.out1.*.phdf"))
+    nfiles = len(filenames)
+    for i in range(nfiles):
+        data = Dump3D(filenames[i],extractvars=params['varname'])
+        iofile=f'TwoDSlice{i:04d}.pkl'
+        print(f"Making {iofile}")
+        slice_data=Make2DSlice(data,sliceaxis=params['sliceaxis'],extractvars=params['varname'])
+        #Saving the slice with python's pickel.  We might consider using hdf5 instead.
+        with open(iofile, "wb") as f:
+            pickle.dump(slice_data, f)
+
+    return
+
+def MakeSlicesVsCoord(params):
+    import pickle
+    #List of outfile names                                                      
+    filename = params['phdffile']
+    data = Dump3D(filename,extractvars=params['varname'])
+    
+    sliceaxis = params['sliceaxis']
+    if (sliceaxis == 1):
+        wmin = np.min(data.zf[:,:-1])
+        wmax = np.max(data.zf[:,:-1])
+    elif (sliceaxis == 2):
+        wmin = np.min(data.yf[:,:-1])
+        wmax = np.max(data.yf[:,:-1])
+    elif (sliceaxis == 3):
+        wmin = np.min(data.xf[:,:-1])
+        wmax = np.max(data.xf[:,:-1])
+
+    nslices = params['nslices']
+    wslices = np.linspace(wmin,wmax,num=nslices)
+    
+    for i in range(nslices):
+        iofile=f'TwoDSlice{i:04d}.pkl'
+        print(f"Making {iofile}")
+        slice_data=Make2DSlice(data,sliceaxis=params['sliceaxis'],slice=wslices[i],extractvars=params['varname'])
+        #Saving the slice with python's pickel.  We might consider using hdf5 instead.
+        with open(iofile, "wb") as f:
+            pickle.dump(slice_data, f)
+
+    return
+
 def Movie2DSlices(varname='p.density',plotlog=True,varbounds=[-16.,-3.]):
     import pickle
     filenames = sorted(glob(f"*TwoDSlice*.pkl"))
@@ -481,7 +527,8 @@ def main():
     parser.add_argument('--Movie1D', action='store_true')
     parser.add_argument('--CalcOneDProfiles', action='store_true')
     #parser.add_argument('--varname', type=str, default='p.density')
-    parser.add_argument('--MakeSlices', action='store_true')
+    parser.add_argument('--MakeSlicesVsTime', action='store_true')
+    parser.add_argument('--MakeSlicesVsCoord', action='store_true')
     parser.add_argument('--Movie2DSlices', action='store_true')
     args= parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


src/fluid/fluid.cpp
```diff
@@ -645,11 +645,11 @@ TaskStatus CalculateFluidSourceTerms(MeshData<Real> *rc, MeshData<Real> *rc_src)
...
-        src(diag::src_terms(0), b, k, j, i) = 0.0;
-        src(diag::src_terms(1), b, k, j, i) = src(c::momentum(0), k, j, i);
-        src(diag::src_terms(2), b, k, j, i) = src(c::momentum(1), k, j, i);
-        src(diag::src_terms(3), b, k, j, i) = src(c::momentum(2), k, j, i);
-        src(diag::src_terms(4), b, k, j, i) = src(c::energy(), k, j, i);
+        src(b, diag::src_terms(0), k, j, i) = 0.0;
+        src(b, diag::src_terms(1), k, j, i) = src(b, c::momentum(0), k, j, i);
+        src(b, diag::src_terms(2), k, j, i) = src(b, c::momentum(1), k, j, i);
+        src(b, diag::src_terms(3), k, j, i) = src(b, c::momentum(2), k, j, i);
+        src(b, diag::src_terms(4), k, j, i) = src(b, c::energy(), k, j, i);
```
Yurlungur, line 652: good catch!

scripts/python/phoebuspy.py
```diff
@@ -59,6 +115,82 @@ def __init__(self, filename):
                 pass
         return
 
+def ReadParameterFile(FileName='Params.dat'):
```
Yurlungur: As I understand it, this is a parser for your script, right? It might be easier/more maintainable to use `pyyaml` and yaml?

scripts/python/phoebuspy.py
```diff
@@ -131,6 +263,525 @@ def MovieGR(data,varname='monopole_gr/rhoadm',anax=None,anay=None,ylim=None,xlim
...
+    #p20, _, _ = binned_statistic(flat_radius, flat_var, statistic=lambda v: np.percentile(v, 20), bins=radedges)
+    #p50, _, _ = binned_statistic(flat_radius, flat_var, statistic=lambda v: np.percentile(v, 50), bins=radedges)
+    #p80, _, _ = binned_statistic(flat_radius, flat_var, statistic=lambda v: np.percentile(v, 80), bins=radedges)
+
+    ## Stack for plotting
+    #varpercentiles = np.vstack([p20, p50, p80])
+    
+    varpercentiles = np.zeros((3,nradbins))
+    varhopper = [[] for _ in range(nradbins)]
+    for im in range(data.NumMB):
```
Yurlungur, line 260: is this really a scatter plot or is it a historgram? If it's the latter, we might want to rename it. Also, I imagine there's some way to get matplotlib to do this automatically

scripts/python/phoebuspy.py
```diff
@@ -131,6 +263,525 @@ def MovieGR(data,varname='monopole_gr/rhoadm',anax=None,anay=None,ylim=None,xlim
...
+    
+    for im in range(data.NumMB):
+        # w[im] shape: (nz+1), (ny+1), or (nx+1) depending on sliceaxis
+        w_local = w[im, :]
+
+        # Find zones where the slice falls between w_local edges
+        mask = (w_local[:-1] <= slice) & (slice < w_local[1:])
+        
+        if (np.any(mask)):
+            assert np.sum(mask) == 1, f"Multiple indices match in meshblock {im}"
```
Yurlungur, line 314: I'm trying to understand the implication of the slicing machinery here. The cell that overlaps with the axis is chosen, is that right?

scripts/python/phoebuspy.py
```diff
@@ -131,6 +263,525 @@ def MovieGR(data,varname='monopole_gr/rhoadm',anax=None,anay=None,ylim=None,xlim
...
+    elif(sliceaxis==2):
+        w=data.yf
+    elif(sliceaxis==3):
+        w=data.xf
+    else:
+        raise ValueError("sliceaxis must be 1 (z), 2 (y), or 3 (x)")
+
+    slice_data = []
+    
+    for im in range(data.NumMB):
```
Yurlungur, line 306: Can this loop be "numpy-ized" so that there doesn't need to be an explicit for loop?

scripts/python/phoebuspy.py
```diff
@@ -131,6 +263,525 @@ def MovieGR(data,varname='monopole_gr/rhoadm',anax=None,anay=None,ylim=None,xlim
...
+            'slice': slice,
+            'meshblock': im,
+            'index': [idx0, idx1],
+            'slice_vars': slice_vars,
+            'coords': coords
+        })
+
+    return slice_data
+
+#def Make1DSlice(data,keepaxis=1,slice=[0.,0.],extractvars=['p.density']):
```
Yurlungur: There is quite a bit of commented out code throughout the file. Do you need/want to keep it all? Looks like quite a bit of it is previous drafts or unused debugging statements, etc.

Yurlungur (approved):
This is a welcome addition! A few nitpicks and suggestions below. But it's all optional, we can merge when you're happy with it.

src/fluid/fluid.cpp
curiousmiah: Thanks ;)

scripts/python/phoebuspy.py
curiousmiah: Good point!  I'm clearly "old school".

scripts/python/phoebuspy.py
curiousmiah: This plots an "average" radial profile for 3D sims.  For example, it plots something like <rho> vs. radius.  It's not really a histogram.  To create this "average" radial profile, I need to calculate some sort of statistic as a function of radius, hence the binning.  Instead of an average, I calculate the median.  But I also plot the 20 and 80th percntiles vs. radius as well.  This will give a sense of median and variation as a function of radius.  While I'm using statistics, I wouldn't call these histograms.  Maybe I could call it CalcStatistics3DvsRadius?

scripts/python/phoebuspy.py
curiousmiah: Maybe.  I was trying, but the conditionals, the construction of the dicts for each block, and the variable sizes of the vars makes it difficult to numpy-ize.  ChatGPT wasn't much help in this regard either.

scripts/python/phoebuspy.py
curiousmiah: Yes.  w_local is an array of the edges for the axis under consideration.  In the case that two overlap the axis -- or more specifically the slice -- then it picks the one where the lower edge is on the axis (or slice).

scripts/python/phoebuspy.py
curiousmiah: Good point.  I'll clean it up!

scripts/python/phoebuspy.py
Yurlungur: sure that works

scripts/python/phoebuspy.py
Yurlungur: gotcha. Yeah that seems reasonable

scripts/python/phoebuspy.py
Yurlungur: ok its fine for now

scripts/python/phoebuspy.py
curiousmiah: Done.

scripts/python/phoebuspy.py
curiousmiah: I just changed the ReadParameterFile file to use yaml.

curiousmiah:
@Yurlungur , I've addressed all of the suggestions. I'm fairly certain that I have adopted all suggestions.  Please check to make sure that you like the direction of the changes.  If so, I'm ready for you to approve merge if all looks good to you as well.

scripts/python/phoebuspy.py
```diff
@@ -1,10 +1,23 @@
...
+#If phoebuspy conda environment is already active
+#conda deactivate
+#If you want to delete the old environment and start from scratch.
+#module unload xalt
+#conda env remove -n phoebuspy
+#If you want to create the phosbuspy conda environment
+#conda create -n phoebuspy numpy numexpr matplotlib h5py scipy pyyaml -y
+#The above needs to be once to define the environment...
+#Do this every time you need to use python on Polaris
+#conda activate phoebuspy
```
Yurlungur, line 19: good to document! :+1:

Yurlungur (approved):
Looks good! Thanks @curiousmiah !

PR #241, opened 2026-03-11
calculate con2prim enthalpy bound

logalexw:
## PR Summary

This is to address a bug I encountered when working on the CCSNe problem; the root find in `con2prim_robust.cpp` isn't bracketed when initializing to the gr grid. This occurs when using a Stellar Collapse EOS (SFHo) for a progenitor at infall.

In [Kastaun et al. 2021](10.1103/PhysRevD.103.023018) there’s a required lower bound (h0) for relativistic enthalpy (h) s.t. 0 < h0 ≤ h. This also serves as the upper bound for the unbracketed root find, i.e. root find using regula falsi from (0, 1/h0].

h0 is currently hardcoded to 1.0, which breaks our upper bound condition if  h < 1, which does happen if you’re using a StellarCollapse EOS (see fig. 1 below), espeically at Ye values that are relevant for SNe. When looking at the evolution of the root find upper bound for CCSNe progenitor, it's visible that the unbracketed root find (red dashed line) occurs when 1/h ≥ 1/h0 1 (see fig. 2 below).  

Goal is to fix with something that's eos-conscious; there's existing primitives in con2prim that should be able to calculate a reasonable bound using singularity-EOS (the existing method in `compute_upper_bound()` isn't sufficient for CCSNe problem). 

Unknown-1

_Fig. 1._ Minimum enthalpy for the SFHo table at relevant electron fraction (Ye) range.

---
Unknown-2

_Fig. 2._ Evolution of upper bound (1/h0) for a CCSNe progenitor, with location of unbracketed root find marked in red. _h_ has been scaled so that the evolution is visible in the same range as the specific internal energy.

## PR Checklist

- [ ] Explain what you did.

### DIFF
--- a/src/fluid/con2prim_robust.hpp
+++ b/src/fluid/con2prim_robust.hpp
@@ -45,6 +45,8 @@ struct FailFlags {
 
 class Residual {
  public:
+
+ // TODO: check if this causes compile error (leaving const variables in constructor def.)
   KOKKOS_FUNCTION
   Residual(const Real D, const Real q, const Real bsq, const Real bsq_rpsq,
            const Real rsq, const Real rbsq, const Real v0sq, const Real Ye,
@@ -57,6 +59,28 @@ class Residual {
     Real garbage = 0.0;
     bounds_.GetFloors(x1_, x2_, x3_, rho_floor_, garbage);
     bounds_.GetCeilings(x1_, x2_, x3_, gam_max_, e_max_);
+
+    rho_floor_ *= floor_scale_fac_;
+  }
+
+  // NEW: overloaded constructor that handles calculation of h0sq, v0sq internally
+  KOKKOS_FUNCTION
+  Residual(const Real D, const Real q, const Real bsq, const Real bsq_rpsq,
+           const Real rsq, const Real rbsq, const Real Ye,
+           const Microphysics::EOS::EOS &eos, const fixup::Bounds &bnds, const Real x1,
+           const Real x2, const Real x3, const Real floor_scale_fac)
+      : D_(D), q_(q), bsq_(bsq), bsq_rpsq_(bsq_rpsq), rsq_(rsq), rbsq_(rbsq),
+        eos_(eos), bounds_(bnds), x1_(x1), x2_(x2), x3_(x3),
+        floor_scale_fac_(floor_scale_fac) {
+
+    lambda_[0] = Ye;
+    Real garbage = 0.0;
+    bounds_.GetFloors(x1_, x2_, x3_, rho_floor_, garbage);
+    bounds_.GetCeilings(x1_, x2_, x3_, gam_max_, e_max_);
+
+    h0sq_ = calc_h0sq();
+    Real zsq_ = rsq_ / h0sq_; // TODO: check that nothing breaks in this normalization.
+    v0sq_ = std::min(zsq_ / (1.0 + zsq_), 1.0 - 1.0 / (gam_max_ * gam_max_));
 
     rho_floor_ *= floor_scale_fac_;
   }
@@ -140,11 +164,18 @@ class Residual {
     return mu - muhat;
   }
 
+  // KOKKOS_INLINE_FUNCTION
+  // Real compute_upper_bound(const Real h0sq) {
+  //   auto func = [=](const Real x) { return aux_func(x, h0sq); };
+  //   root_find::RootFind root;
+  //   return root.itp(func, 1.e-16, 1.0 / sqrt(h0sq), 1.e-3, -1.0);
+  // }
+
   KOKKOS_INLINE_FUNCTION
-  Real compute_upper_bound(const Real h0sq) {
-    auto func = [=](const Real x) { return aux_func(x, h0sq); };
+  Real compute_upper_bound() {
+    auto func = [=](const Real x) { return aux_func(x); };
     root_find::RootFind root;
-    return root.itp(func, 1.e-16, 1.0 / sqrt(h0sq), 1.e-3, -1.0);
+    return root.itp(func, 1.e-16, 1.0 / sqrt(h0sq_), 1.e-3, -1.0);
   }
 
   KOKKOS_INLINE_FUNCTION
@@ -162,7 +193,9 @@ class Residual {
   }
 
  private:
-  const Real D_, q_, bsq_, bsq_rpsq_, rsq_, rbsq_, v0sq_;
+  // const Real D_, q_, bsq_, bsq_rpsq_, rsq_, rbsq_, v0sq_;
+  const Real D_, q_, bsq_, bsq_rpsq_, rsq_, rbsq_;
+  Real h0sq_, v0sq_; // these cannot be const, we need to set them after initialization.
   const Microphysics::EOS::EOS &eos_;
   const fixup::Bounds bounds_;
   const Real x1_, x2_, x3_;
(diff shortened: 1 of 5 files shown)


Yurlungur:
Good catch! Should have known we would need to worry about this--nubhlight also has to handle this case, as does the torus initial conditions.

c-prather:
Oh, I would benefit from stealing this too I think. Though iirc I set the brackets following AthenaK's implementation, so maybe I ended up with an enthalpy bound from that.
It might be worth your time looking at KHARMA's [stolen & modded](https://github.com/AFD-Illinois/kharma/blob/stable/kharma/inverter/kastaun.hpp) version of this function, we have some downstream tweaks that really improve stability.

Yurlungur:
@logalexw make sure to remove "WIP" and ping when you're ready for review. No rush though!

logalexw:
I've refactored the original fix to add an parameter that stors the enthalpy lower bound and is able to be utilized in conserved to primitive reconstruction (both initialization and fixup)!

Unfortunately, there seems to be another bug (absolute value??) that causes epsilon to be strictly positive. The new updates to `con2prim_robust.cpp` produce the dashed line, which approximates our expected initial SIE well, but phoebus produces the "mirrored" SIE curve (in blue) on the zeroth timestep. Any ideas where this is happening? I've had no luck finding the source in `src/fixup` so far, which is what @AstroBarker originally suspected.

---
image

Yurlungur:
Is epsilon here specific internal energy? It's like the floors.

Yurlungur:
An energy floor is an optional feature in phoebus. You can try disabling or making the floor very small in the input deck.

logalexw:
I thought it was the floors as well, I've set them to a reasonable negative value (for the SC EOS tables + CCSNe problem)– I was debugging with verbose output and the floors (sie, u, rho) were all at expected values (SFHo table minimum for sie, very small values for rho, etc.)

Yurlungur:
Ah ok. Well if it's not floors, the next thing I would check would just to grep through the code for calls to `std::abs` or `fabs`. There will be a lot... but hopefully you can quickly narrow it down. I agree an absolute value seems like a suspect.

AstroBarker:
Good work @logalexw ! When you have time could you share your workflow to reproduce this (cmake build line for example). And this flipped behavior is present in the `00000.phdf` output? And you're plotting the `sie` output directly? And your fixup block looks like this:
```
<fixup>
enable_floors = true
enable_ceilings = true
enable_flux_fixup = true
enable_c2p_fixup = true
floor_type = ConstantRhoSie
# updated floors!
rho0_floor = 1.0e-10
sie0_floor = -1.0e2 # << changed back to original oom.
ceiling_type = ConstantGamSie
sie0_ceiling = 1.0e3
gam0_ceiling = 1.0e4
```
I am asking as I am having some trouble reproducing the behavior using this branch.

AstroBarker:
And I am seeing that the docs action is failing with a `sphinx-multiversion` issue.. This thing is so unstable, they changed the version of something under the hood in runner. Happened to me on another project. I'll check for the fix before we merge.

Yurlungur:
> And I am seeing that the docs action is failing with a `sphinx-multiversion` issue.. This thing is so unstable, they changed the version of something under the hood in runner. Happened to me on another project. I'll check for the fix before we merge.

Yeah just need to pin sphinx. Here's the fix in singularity-eos
https://github.com/lanl/singularity-eos/blob/main/.github/workflows/docs.yml#L34

logalexw:
> Good work @logalexw ! When you have time could you share your workflow to reproduce this (cmake build line for example). And this flipped behavior is present in the `00000.phdf` output? And you're plotting the `sie` output directly? And your fixup block looks like this:
> 
> ```
> <fixup>
> enable_floors = true
> enable_ceilings = true
> enable_flux_fixup = true
> enable_c2p_fixup = true
> floor_type = ConstantRhoSie
> # updated floors!
> rho0_floor = 1.0e-10
> sie0_floor = -1.0e2 # << changed back to original oom.
> ceiling_type = ConstantGamSie
> sie0_ceiling = 1.0e3
> gam0_ceiling = 1.0e4
> ```
> 
> I am asking as I am having some trouble reproducing the behavior using this branch.

Yes, these are the correct parameters, and yes, the flip happens in `00000.phdf` - I get the same result plotting both `sie` directly and using `p.energy/p.density`.  

I've built this on UA HPC using `cmake .. -DPHOEBUS_GEOMETRY=MonopoleSph; make -j8`

logalexw:
Another thing I just saw - is it intended that both `drho`, `du` and other derived quantities are always positive in `fixup.cpp`? Allowing for negative energy floors breaks that assumption.

Yurlungur:
> Another thing I just saw - is it intended that both `drho`, `du` and other derived quantities are always positive in `fixup.cpp`? Allowing for negative energy floors breaks that assumption.

I'm not sure I understand... shouldn't they still be always positive? If u < uflr, and |u| > |uflr|, then |du| > 0 still to add energy, right?

logalexw:
previous issue with mirrored sie is resolved (thanks to @AstroBarker for the catch)!

src/fluid/con2prim_robust.hpp
```diff
@@ -360,19 +383,20 @@ class ConToPrim {
...
                  floor_scale_fac_);
 
-    // find the upper bound
-    // TODO(JCD): revisit this.  is it worth it to find the upper bound?
-    //            Doesn't seem to be at a quick glance.
-    // const Real mu_r = res.compute_upper_bound(h0sq_);
-    // solve
+    // conditional from Kastaun et al. 2021, we need a tighter upper bound if r > h0 due
+    // to sharp kink in lorentz factor.
+    Real mu_r;
```
AstroBarker, line 392: Purely stylistic, but I would add braces {} to the if/else here.

src/microphysics/eos_phoebus/eos_phoebus.cpp
```diff
@@ -25,6 +25,9 @@
 #include "phoebus_utils/unit_conversions.hpp"
 #include "phoebus_utils/variables.hpp"
 
+// inspired by SPACELOOP in geometry utils, util for global lower bound for enthalpy
```
AstroBarker: I would suggest either removing this and writing the for loops out explicitly below or moving it to a better home, maybe somewhere under `phoebus_utils` and including the relevant header in this file. Either is fine.

AstroBarker (approved):
Nice work fixing this! Looks good, a couple of small suggestions for mostly cleanliness and organization.

src/microphysics/eos_phoebus/eos_phoebus.cpp
Yurlungur: I think it's fine here... It's local to this implementation file, so it can clearly be seen what it's used for. That said, the loops in geometry are written to convey intent... the bounds are always the same. If this macro is to be generalized it needs a more useful name than LOOP, like LOOP1D or something. Or PHOEBUSFOR or similar. 

Personally I think this isn't really saving us enough characters that I feel the need to factor it out to utils.

src/microphysics/eos_phoebus/eos_phoebus.cpp
```diff
@@ -154,6 +162,35 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+        LOOP(t, n) {
+          eps = eos_sc.InternalEnergyFromDensityTemperature(rho, T, lambda) / sie_unit;
+          P = eos_sc.PressureFromDensityTemperature(rho, T, lambda) / press_unit;
+          h_min = std::min(h_min, 1 + eps + robust::ratio(P, rho));
+          T += dT;
+        }
+        rho += drho;
+      }
+      ye += dye;
+    }
```
Yurlungur: This is fine for now but I *think* it can probably be done better... For example, one could do a Newton method to find the point of minimal enthalpy... Or by having the EOS report this itself internally. Food for thought for later.

src/microphysics/eos_phoebus/eos_phoebus.cpp
```diff
@@ -154,6 +162,35 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+    eps = 0.0;
+
+    // spacing for each dimension (i.e. np.linspace)
+    dT = (T_max - T_min) / n;
+    drho = (rho_max - rho_min) / n;
+    dye = (ye_max - ye_min) / (n / 2); // we don't need to resolve ye as much
+    // initial conditions
+    T = T_min;
+    rho = rho_min;
+    ye = ye_min;
```
Yurlungur: The table spans many orders of magnitude. I think you can use a coarser spacing if you used a log spacing in density and temperature. Or, realistically, you can simply only look at relatively low densities and temperatures, as high densities and temperatures simply won't have a small enthalpy.

Yurlungur (approved):
Nice work! I have a few comments about the enthalpy finding loop but my comments are non-blocking.

src/microphysics/eos_phoebus/eos_phoebus.cpp
logalexw: I agree with Brandon; the macro originally had more functionality before I simplified the logic, but in this form the intent would probably be just as clear as a set of explicit for loops (and it's not really saving _that_ much room).

src/microphysics/eos_phoebus/eos_phoebus.cpp
logalexw: @carlnotsagan and I were talking about this as a later wishlist item for the EOS as well. I'm definitely open to doing something more efficient in the future as well, i don't think this is the best fix per se.

Yurlungur: Performancewise it probably doesn't matter, since it's just initial conditions. But I worry it's a bit brittle.

logalexw:
final formatting updates from the comments above are complete! if there are no further adjustments/suggestions, this should be ready to merge.

Yurlungur (approved):

Yurlungur:
Merged. Thanks for the fix @logalexw !

PR #239, opened 2025-08-01
Fixed logic errors in history.cpp and progenitordata.cpp that caused …

graveltr:
…failures for runs with radiation turned off.

## PR Summary

Checks of whether radiation is active were not being done when computing various gain region quantities. This caused failures, for example, for homologous.pin. This PR implements those checks and leads to homologous.pin running without issue.

## PR Checklist

- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Format your changes by calling `scripts/bash/format.sh`.
- [x] Explain what you did.
- [ ] Make any necessary changes to the documentation.

PR #243, opened 2026-04-30
fix logic for gain calculations + ye bug

logalexw:
## PR Summary

Found a small bug when working on the CCSNe problem regarding logic for gain region calculations— if you turn radiation on (`rad = true`) and use any method other than `method = cooling_function` (e.g. lightbulb), phoebus crashes. 

Looks like the parameter `do_gain_reducer` is only initialized and defined if `do_lightbulb` is true, but `history.cpp` expects this param to exist if radiation is turned on (independent of method). @AstroBarker mentioned that most of this logic was fixed in #239 but this was an edge case that was missed. 

Currently thinking the most comprehensive fix is a param under `<radiation>` in the input files that turns gain calculations on or off (defaulting to false) so that sure relevant parameters are defined outside of the lightbulb conditionals. Suggestions for logic/implementation welcome!

**edit:** also added a small update to fix a bug in the electron fraction evolution in `src/fixup/fixup.cpp` that we found where a primitive value was being set to a conserved one instead.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/inputs/progenitor.pin
+++ b/inputs/progenitor.pin
@@ -139,7 +139,7 @@ cfl = 0.5
 method = cooling_function
 do_lightbulb 	= true
 do_liebendorfer	= true
-do_gain_calc = false # can be true or false for lightbulb only; always false for other methods
+do_gain_calc = false # defaults to false if not set
 
 <monopole_gr>
 enabled = true
--- a/inputs/progenitor.pin
+++ b/inputs/progenitor.pin
@@ -132,7 +132,14 @@ ceiling_type = ConstantGamSie
 sie0_ceiling = 10.0;
 gam0_ceiling = 10.0;
 
-
+# rad disabled in default *.pin; see <physics> above
+# simple example for lightbulb case
+<radiation> 
+cfl = 0.5 
+method = cooling_function
+do_lightbulb 	= true
+do_liebendorfer	= true
+do_gain_calc = false # can be true or false for lightbulb only; always false for other methods
 
 <monopole_gr>
 enabled = true
--- a/src/analysis/history.cpp
+++ b/src/analysis/history.cpp
@@ -306,8 +306,9 @@ Real CalculateMdot(MeshData<Real> *md, Real rc, bool gain) {
 
   auto rad = pmb->packages.Get("radiation").get();
   const bool rad_active = rad->Param<bool>("active");
+  const bool do_gain_calc = rad->Param<bool>("do_gain_calc");
   bool do_gain = false;
-  if (rad_active) {
+  if (rad_active && do_gain_calc) {
     const parthenon::AllReduce<bool> *pdo_gain_reducer =
         rad->MutableParam<parthenon::AllReduce<bool>>("do_gain_reducer");
     do_gain = pdo_gain_reducer->val;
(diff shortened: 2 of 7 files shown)


logalexw:
Logic has been updated to include a `do_gain_calc` parameter that is false by default and pairs with the other gain-related parameters that are used for calculations in the lightbulb case! I also updated the original `progenitor.pin` deck to include an example for the lightbulb case for future clarity.

(the ye bug in `fixup.cpp` is also fixed.)

src/analysis/history.cpp
```diff
@@ -306,8 +306,10 @@ Real CalculateMdot(MeshData<Real> *md, Real rc, bool gain) {
 
   auto rad = pmb->packages.Get("radiation").get();
   const bool rad_active = rad->Param<bool>("active");
+  const bool do_gain_calc = rad->Param<bool>("do_gain_calc");
   bool do_gain = false;
-  if (rad_active) {
+
+  if (rad_active && do_gain_calc) {
```
AstroBarker, line 312: I think that that `&&` is redundant here -- `do_gain_calc` should be defaulted to false when radiation is disabled.

src/radiation/radiation.cpp
```diff
@@ -113,6 +113,21 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   bool absorption = pin->GetOrAddBoolean("radiation", "absorption", true);
   params.Add("absorption", absorption);
 
+  // initialize gain parameters from .pin, set false by default
+  bool do_gain_calc = pin->GetOrAddBoolean("radiation", "do_gain_calc", false);
+  params.Add("do_gain_calc", do_gain_calc, false);
+
+  // only intitialize these if we do gain calculations; otherwise not needed!
+  if (do_gain_calc) {
+    printf("do_gain_calc true\n\n");
```
AstroBarker: Oops

inputs/progenitor.pin
```diff
@@ -132,7 +132,14 @@ ceiling_type = ConstantGamSie
...
 
-
+# rad disabled in default *.pin; see <physics> above
+# simple example for lightbulb case
+<radiation> 
+cfl = 0.5 
+method = cooling_function
+do_lightbulb 	= true
+do_liebendorfer	= true
+do_gain_calc = false # can be true or false for lightbulb only; always false for other methods
```
AstroBarker: Can you elaborate on "always false for other methods"?

AstroBarker (approved):
This looks good @logalexw ! Thanks for the cleaning, including fixing my sin. Just a few questions/requests before merging.

src/analysis/history.cpp
logalexw: Good point– however, the param that's called in that logical (`do_gain_reducer`) only exists if `do_gain_calc = true` with the current logic in radiation (which could be tweaked)

inputs/progenitor.pin
logalexw: Good catch, I've made this clearer now :)

logalexw:
Clarifications and last minor tweaks are complete! This should be ready for merge now @AstroBarker, thanks for the review and suggestions.

edit: the new params do work as expected with radiation active and turned off (yay)

Yurlungur (approved):

PR #244, opened 2026-06-09
adding density-temperature floors

logalexw:
## PR Summary

Realized on my quest to get CCSNe working in phoebus that it might be a good idea to have floors that weren't dependent on inherently negative values, e.g. negative specific internal energies from tabulated equations of state (ref. #241 for additional context).

Adding a floor that relies on constant density and temperature seemed like a reasonable step given other supernovae codes (like GR1D) and a pretty lightweight addition. This is modeled fairly closely to the current implementation of `ConstantRhoSie`, with some modifications made to quantities that are updated in `fixup.cpp` once the conditions for floors to be applied are met. Suggestions for science/implementation are welcome and appreciated!

**bonus.** also found a minor logic bug that slipped through in #243 that is now fixed (things broke for the homologous case, oops).

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/fixup/fixup.cpp
+++ b/src/fixup/fixup.cpp
@@ -140,40 +140,40 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
       Real rho0 = pin->GetOrAddReal("fixup", "rho0_floor", 0.0);
       Real sie0 = pin->GetOrAddReal("fixup", "sie0_floor", 0.0);
       params.Add("floor", Floors(constant_rho_sie_floor_tag, rho0, sie0));
-      params.Add("is_rhoT_floor", false);
+      params.Add("is_floor_rhoT", false);
     } else if (floor_type == "ExpX1RhoSie") {
       Real rho0 = pin->GetOrAddReal("fixup", "rho0_floor", 0.0);
       Real sie0 = pin->GetOrAddReal("fixup", "sie0_floor", 0.0);
       Real rp = pin->GetOrAddReal("fixup", "rho_exp_floor", -2.0);
       Real sp = pin->GetOrAddReal("fixup", "sie_exp_floor", -1.0);
       params.Add("floor", Floors(exp_x1_rho_sie_floor_tag, rho0, sie0, rp, sp));
-      params.Add("is_rhoT_floor", false);
+      params.Add("is_floor_rhoT", false);
     } else if (floor_type == "ExpX1RhoU") {
       Real rho0 = pin->GetOrAddReal("fixup", "rho0_floor", 0.0);
       Real sie0 = pin->GetOrAddReal("fixup", "u0_floor", 0.0);
       Real rp = pin->GetOrAddReal("fixup", "rho_exp_floor", -2.0);
       Real sp = pin->GetOrAddReal("fixup", "u_exp_floor", -3.0);
       params.Add("floor", Floors(exp_x1_rho_u_floor_tag, rho0, sie0, rp, sp));
-      params.Add("is_rhoT_floor", false);
+      params.Add("is_floor_rhoT", false);
     } else if (floor_type == "X1RhoSie") {
       Real rho0 = pin->GetOrAddReal("fixup", "rho0_floor", 0.0);
       Real sie0 = pin->GetOrAddReal("fixup", "u0_floor", 0.0);
       Real rp = pin->GetOrAddReal("fixup", "rho_exp_floor", -2.0);
       Real sp = pin->GetOrAddReal("fixup", "u_exp_floor", -3.0);
       params.Add("floor", Floors(x1_rho_sie_floor_tag, rho0, sie0, rp, sp));
-      params.Add("is_rhoT_floor", false);
+      params.Add("is_floor_rhoT", false);
     } else if (floor_type == "RRhoSie") {
       Real rho0 = pin->GetOrAddReal("fixup", "rho0_floor", 0.0);
       Real sie0 = pin->GetOrAddReal("fixup", "u0_floor", 0.0);
       Real rp = pin->GetOrAddReal("fixup", "rho_exp_floor", -2.0);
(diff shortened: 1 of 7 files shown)


src/fixup/fixup.cpp
```diff
@@ -164,6 +164,11 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
       Real rp = pin->GetOrAddReal("fixup", "rho_exp_floor", -2.0);
       Real sp = pin->GetOrAddReal("fixup", "u_exp_floor", -3.0);
       params.Add("floor", Floors(r_rho_sie_floor_tag, rho0, sie0, rp, sp));
+    } else if (floor_type == "ConstantRhoTemp") {
+      Real rho0 = pin->GetOrAddReal("fixup", "rho0_floor", 0.0);
+      Real sie0 = pin->GetOrAddReal("fixup", "sie0_floor", 0.0);
```
Yurlungur: Why is sie0 necessary for constant rho/T?

src/fixup/fixup.cpp
```diff
@@ -434,11 +444,19 @@ TaskStatus ApplyFloorsImpl(T *rc, IndexDomain domain = IndexDomain::entire) {
...
           floor_applied = true;
           drho = std::max<Real>(drho, du / sie_floor);
           du = std::max<Real>(du, sie_floor * drho);
+        } else if (drho > 0. || dT > 0. && stellar_collapse) {
+          floor_applied = true;
+          // new, rho-T consistent energy density (temp floor dependent?)
+          du = std::max<Real>(du, eos.InternalEnergyFromDensityTemperature(
+                                      rho_floor_max, T_floor_max, eos_lambda) *
+                                          rho_floor_max -
+                                      v(b, peng, k, j, i));
```
Yurlungur: Why branch like this? All EOS's support the sie(rho,T) call?

src/pgen/torus.cpp
```diff
@@ -340,8 +340,8 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
         // fixup
         Real rhoflr = 0;
-        Real epsflr;
-        floor.GetFloors(x1v, x2v, x3, rhoflr, epsflr);
+        Real epsflr, garbage;
+        floor.GetFloors(x1v, x2v, x3, rhoflr, epsflr, garbage);
```
Yurlungur: The temp floor should also work for the torus, so it might be worth trying it.

Yurlungur (commented):
This is a good idea! And it indeed may help. But why branch on stellarcollapse? Temperature floors should work for any singularity-eos model. And for ideal gas, they're equivalent to sie floors, up to a constant.

src/fixup/fixup.cpp
logalexw: Good point, I can just have the sie floor pull from the EOS bounds since it's still used in some of the con2prim methods.

src/fixup/fixup.cpp
logalexw: Hm, maybe it makes more sense to have a branch just for the const. rho-T flag instead. I assumed that the tabulated stellar collapse eos would be the only one to run into an issue with negative sie/u floors, but that might be more restrictive/case-specific than needed.

Yurlungur: In Phoebus that is probably true, but I don't like special-casing like this. And there might be reasons to use a constant rho-T floor in other contexts.

src/fixup/fixup.cpp
Yurlungur: Ah I see. Yes, I'd pull the sie floor by evaluating sie(rho, Tfloor). Note it will be different per rho.

If rho isn't available, then you will need to compute min_rho (sie(rho, T_floor))

logalexw:
Added some implementation for the const. density-temperature floor in the torus case, also included some local (e.g. ye-dependent) limits for the sie floor in con2prim for cases where the `ConstRhoTemp` floor is used but a minimum energy is needed.

logalexw:
**edit.** also found a minor logic bug that was missed on #243 - now fixed!

src/fixup/fixup.cpp
```diff
@@ -434,11 +448,19 @@ TaskStatus ApplyFloorsImpl(T *rc, IndexDomain domain = IndexDomain::entire) {
...
+        Real dT = T_floor_max - v(b, tmp, k, j, i);
 
-        if (drho > 0. || du > 0.) {
+        if (drho > 0. || du > 0. && !is_floor_rhoT) {
           floor_applied = true;
           drho = std::max<Real>(drho, du / sie_floor);
           du = std::max<Real>(du, sie_floor * drho);
+        } else if (drho > 0. || dT > 0. && is_floor_rhoT) {
+          floor_applied = true;
+          // new, rho-T consistent energy density (temp floor dependent)
```
AstroBarker: Minor style nitpick, but now we can remove reference to this being "new" and just describe the energy density choice

src/fluid/con2prim_robust.hpp
```diff
@@ -105,8 +105,13 @@ class Residual {
   Real ehat_mu(const Real mu, const Real qbar, const Real rbarsq, const Real vhatsq,
                const Real What) {
     Real rho = rhohat_mu(robust::ratio(1.0, What));
-    bounds_.GetFloors(x1_, x2_, x3_, rho, e_floor_);
+    bounds_.GetFloors(x1_, x2_, x3_, rho, e_floor_, temp_floor_);
     e_floor_ *= floor_scale_fac_;
+    Real e_floor_local_ =
```
AstroBarker: `e_floor_local` can be made `const` (`rho` above too, while I am looking here).

AstroBarker (approved):
Looks like a good addition and sounds like it has solved some problems. Runs on torus. A coupling of small changes, will merge after. Thanks @logalexw !

AstroBarker:
Fixed my mistaken `const` suggestion, hitting the button when tests pass.

PR #245, opened 2026-06-19
refactor of the ccsne/progenitor pipeline

logalexw:
## PR Summary

Working on some cleanup/refactoring of the python pipeline that takes ccsne progenitors (KEPLER, MESA, GR1D) as input and maps them to a phoebus-like grid + calcuates ADM quantities. Adding a few new features for interpolation type, grid formatting, user interface, and final i/o as well.

There's also a few minor tweaks in the main codebase for consistency (mainly in `pgen/progenitor.cpp`), but no major changes outside of scripts.

TBD for additional updates + description as this progresses-- suggestions and discussion always welcome!

## PR Checklist

- [x] Explain what you did.


logalexw:
Update: the refactor is (finally) done! The new pipeline (`scripts/python/ccsne`) takes @mari2895's original `grsolver.py` + ccsne processing scripts and preserves the original calculations while adding some user-facing improvements and updated computation methods. New features include:

1. **Simplified interface** - it now only takes one command to run the entire pipeline (vs. independently running several scripts)! Users can build their own input deck (similar to the main `phoebus` structure) with a variety of different options for progenitor pre-processing, ADM calculation methods, and i/o. 
2. **Documentation** - there's some minor updates to the Sphinx configurations + an entirely new page for the CCSNe API that builds autodocs from the docstrings in the new module. I've also included some details + references on the math and GR happening in the solver to make it clearer for future reference.
3. **Progenitor processing** - all of the separate scripts to handle different stellar progenitors (`MESA`, `KEPLER`, `GR1D`, etc.) are now in one unified workflow, with some useful features for data retrieval and progenitor pre-processing (i.e. increasing from Lagrangian to Eulerian-appropriate resolution, smoothing data, fixup)
4. **Nicer I/O** - all files output from the pipeline (original profiles, post-ADM profiles, `phoebus` input files) are now human-readable with a uniform (and informative) naming convention. I've also added the option to generate a `*.info` file that contains helpful values for building an input deck, like code unit conversions, characteristic values, and bounds of the EOS and progenitor profile.
5. **New interpolation** - the original method used for calculation and interpolation of ADM quantities in `grsolver.py` is now deprecated (`scipy.interpolation.interp1d`), so I've added an implementation of the same piecewise quadratic interpolation @Yurlungur uses for the main ADM calculations + interpolation in `src/monopole_gr/monopole_gr_base.hpp` as an alternative option.

If there are questions or suggestions regarding this update, please let me know! I've included a few figures from the V&V runs I did below to show consistency/improvement for both primitive and ADM calculations, and can provide quite a bit more data at request.

---

**Note.** all figures here are for a 15 Msun `MESA` model, but the pipeline has also been tested and verified for `KEPLER` and `GR1D` inputs (although I think there's one input that might need to be adjusted for the latter later on; it should only be an issue if one is running the pipeline with a Helmholtz EOS instead of StellarCollapse).

The black dashed line is always the original `grsolver.py` output in each case. All figures also include a panel with relative error in comparison to the original `grsolver.py` values.

### direct comparisons, deprecated methods

**Fig. 1.** Comparison of initial, pre-ADM progenitor velocity profile. 
download

**Fig. 2.** Comparison of post-ADM progenitor density profiles. Note that the cubic + piecewise profile uses cubic interpolation for primitives and piecewise interpolation for ADM calculations. Note that clamped b.c. are _not_ recommended if using cubic interpolation.
download

**Fig. 3.** Same as above, but for ADM density profile.
download

### direct comparisons, new sampling + methods

Finally, we do some comparisons with the original pipeline and what I would use for a new production run input, with the original progenitor included in yellow.

**Fig. 4.** Comparison of post-ADM progenitor density profile. 
download

**Fig. 5.** Same as above, but for ADM density profile.
download

scripts/python/ccsne/adm.py
```diff
@@ -0,0 +1,704 @@
...
+                + 1.0 / (a[i] ** 2.0 * r[i] * eps)
+            )
+            M[i, i - 1] = A
+            M[i, i] = B
+            M[i, i + 1] = C
+
+        def f(x):
+            return np.dot(M, x) + V
+
+        sol = optimize.root(f, 100 * np.ones(self.zones))
```
AstroBarker: On quick inspection, the matrix is tridiagonal? A dedicated tridiagonal solver from scipy might accelerate this. Not necessarily worth it now but a todo or note could be nice if this script's runtime becomes a nuisance.

scripts/python/ccsne/adm.py
```diff
@@ -0,0 +1,704 @@
...
+        ye          = prof[:, 4]
+        temp        = prof[:, 5]
+        eps         = prof[:, 6]
+        omega       = prof[:, 7]
+        abar        = prof[:, 8]
+        zbar        = prof[:, 9]
+
+        # now we make the new phoebus grid, depending on the given criteria
+        # all grids are uniform and do NOT go to zero (for now), extrapolation occurs after ADM calculations
+        if use_rho_cut:
```
AstroBarker: It looks like the bools here are all optional? What happens when they are all `False` and `self.grid` is not set?

scripts/python/ccsne/adm.py
```diff
@@ -0,0 +1,704 @@
...
+        elif use_custom:
+            self.grid = np.linspace(custom_min, custom_max, self.zones)
+        elif use_def_rad:
+            self.grid = self.r0
+            self.zones = self.r0.shape[0] 
+        
+        # finding energy density (consistent with eos) since we don't include that in profiles:
+        if self.problem == "stellartable":
+         
+            if self.eos_type.lower() == 'helmholtz':
```
AstroBarker: This is fine for now but it might be more future proof to say, e.g., 
```python
if self.eos_type.lower() == 'helmholtz':
  # ...
else if self.eos_type.lower() == 'stellarcollapse': # or whatever this should read
  # ...
else:
  # throw an error: eos {self.eos_type.lower()} now supported for ccsn init
```
Somewhere, at least, it should be protected that a reasonable EOS is used.

scripts/python/ccsne/adm.py
```diff
@@ -0,0 +1,704 @@
...
+
+    '''
+
+    def __init__(self, problem: str, DATPATH: str, EOSPATH: str, eos_type: str = 'stellarcollapse', use_def_rad: bool = False, use_rho_cut: bool = False, rho_cut: float = 2.0e3, use_rad_cut: bool = False, rad_cut: float = 1e9, use_custom: bool = False, custom_min: float = 5e5, custom_max: float = 5e9, zones: int = 10000, interp_method: str = 'cubic', bc_type: str = 'clamped', interp_method_adm: str = 'piecewise', num_iterations: int = 100, dalpha_eps: float = 1.0e-12, extrapolate: bool = True, verbose: bool = True):
+        
+        r'''
+        Constructor for the ADMSolver class.
+
+        Args:
+            problem (str): Which problem you want to generate ``phoebus`` input for. Options are 'StellarTable', 'TOV' (not implemented), and 'homologous' (not implemented).
```
AstroBarker: Since `tov` is not yet implemented, it might be worth throwing in either an early warning that the results might be untrustworthy or an early exit/failure that it is not yet implemented.

scripts/python/ccsne/adm.py
```diff
@@ -0,0 +1,704 @@
...
+        V0 = [1, 0]
+        
+        ## SOLVE COUPLED DIFFERENTIAL EQUATIONS FOR EXTRINSIC CURVATURE
+
+        sol = solve_ivp(f, [r[0], r[-1]], V0, t_eval=r, vectorized=True)
+        result = sol.y
+        
+        # break if original data is too sparse and our ivp dimensions are incorrect...
+        if (result.shape[1] != self.zones):
+            raise ArithmeticError( f'initial raw profile at `{self.DATPATH}` too radially sparse for interpolation. increase the factor or decrease drad spacing and try again.')
```
AstroBarker: Minor but I would probably use a `ValueError` here.

scripts/python/ccsne/adm.py
```diff
@@ -0,0 +1,704 @@
...
+
+        for i in range( self.n ):
+
+            a, K, alpha, beta = self.calculate_metric(rho_adm, P_adm, S_adm)
+
+            if self.verbose: # TODO: add a more informative print statement here
+                print( f'\titeration {i:2d}\talpha = {np.max(abs(alpha)):4.5e}\tdalpha = {np.max(abs(alpha_prev - alpha)):4.5e}')
+
+            # default criteria is 1e-12, loosening to 6e-12 is robust as well.
+            if np.max(abs(alpha_prev - alpha)) < self.dalpha_eps:
```
AstroBarker: Is it possible here for this convergence criteria to be immediately satisfied? In that case, `Srr_adm` would not yet be defined.

scripts/python/ccsne/ccsne.py
```diff
@@ -0,0 +1,209 @@
...
+    
+        else:
+            prof_raw, time = get_GR1D_profile( params.model_path, False, params.timestamp, verbose=params.verbose)
+
+        params.timestamp = time
+
+    # we leave the resolution and primitives as-is for GR1D models since they're evolved.
+    if params.model_type.lower() != 'gr1d':
+        # default
+        if not params.depr:
```
AstroBarker: The old sampling is probably worth removing at some point now that the figures have been made. Can be a separate PR.

scripts/python/ccsne/ccsne.py
```diff
@@ -0,0 +1,209 @@
...
+    parser.add_argument('--wma-all', action = 'store_true', default=False)
+    parser.add_argument('--wma-vel', action = 'store_true', default=False)
+    parser.add_argument('--wma-n', type=int, default=100)
+    parser.add_argument('--wma-exp', type=int, default=3)
+
+    parser.add_argument('--factor-interp', type=int, default=10)
+    parser.add_argument('--use-drad-interp', action = 'store_true', default=False)
+    parser.add_argument('--drad-interp', type=float, default=1e7)
+    
+    parser.add_argument('--use-uniform', action = 'store_true', default=False)
```
AstroBarker: Are `use-uniform` and `unif-max` used below?

doc/sphinx/src/pipeline.rst
```diff
@@ -0,0 +1,230 @@
...
+     - Absolute or relative path to input model/data.
+   * - ``‑‑model‑name``
+     - ``aag21``
+     - Desired name/id for your model.
+   * - ``‑‑model‑type``
+     - ``MESA``
+     - Type of model (options: MESA, GR1D, KEPLER).
+   * - ``‑‑model‑header``
+     - ``-1``
+     - (optional) Line number (zero-indexed) of the profile header (if < 0, goes to defaults - MESA: 4, KEPLER: 1).
```
AstroBarker, line 46: I think, based on usage, the description should read `<= 0` as the check is, for example:
```python
        if params.model_header > 0:
            prof_raw = get_MESA_profile( params.model_path, params.model_header, params.verbose)
        else:
            prof_raw = get_MESA_profile( params.model_path, verbose=params.verbose)
```
or the checks need adjusting.

AstroBarker:
Thanks @logalexw ! The plots look great and this is a lot of effort. I've made several comments for discussion/change. It's a pretty big diff so I haven't looked as closely as I might like, but things seem to work.

AstroBarker:
Also, @logalexw, I am noticing that our CI does not check python formatting. I noticed a lot of white space throughout. Have you passed this through [black](https://pypi.org/project/black/) for formatting?

AstroBarker:
> Also, @logalexw, I am noticing that our CI does not check python formatting. I noticed a lot of white space throughout. Have you passed this through [black](https://pypi.org/project/black/) for formatting?

Looks like we never really formatted any of the other python either, and I don't like black's defaults anyway, so we can put this off for a formatting PR, I think,

doc/sphinx/src/pipeline.rst
```diff
@@ -0,0 +1,230 @@
+.. _singularity-eos: https://lanl.github.io/singularity-eos
+
+The CCSNe API
```
AstroBarker: The title for this page is a bit vague -- it is meant specifically to describe initialization? Then maybe "CCSN Initialization". Or if it is meant to describe the full workflow form initialization through the end product, I would still rename it as something like "The CCSN Workflow" instead of API.

doc/sphinx/src/pipeline.rst
```diff
@@ -0,0 +1,230 @@
...
+==================
+
+.. note::
+
+   These docs (and this pipeline!) are under active development. If you encounter a bug or add a new feature, let the developers know!
+
+If you're interested in exploring the core-collapse supernovae (e.g. ``progenitor`` ) problem in ``Phoebus``,
```
AstroBarker: Nit picky, it should read "supernova problem" (singular).

AstroBarker (changes requested):
Mostly minor suggestions, thanks @logalexw !

doc/sphinx/src/pipeline.rst
logalexw: thanks for the catch! the check itself needed a small adjustment.

scripts/python/ccsne/ccsne.py
logalexw: they should have been... (and are now, thanks)

scripts/python/ccsne/adm.py
logalexw: added! there's now a catch-all user warning as well.

scripts/python/ccsne/adm.py
logalexw: similar to `tov`, there's now better logic and a user warning here too for general cases/user input.

scripts/python/ccsne/adm.py
logalexw: good catch - there's now a small logic fix in the main driver that accounts for that option, as well as defaults added to the method arguments just in case.

scripts/python/ccsne/adm.py
logalexw: mmmm good point. i've added a default initialization outside of the loop/logic on the low chance that occurs.

AstroBarker:
Just a couple new comments @logalexw and I am happy to merge! Thanks for your patience (it is a lot of code you've written for this!).

scripts/python/stellar/stellar.py
```diff
@@ -0,0 +1,211 @@
...
+    parser.add_argument('-f', '--file', type=str, help='name of the input parameter file + path (if needed)', action = LoadFile)
+    parser.add_argument('-v', '--verbose', action = 'store_true')
+
+    # get our input file parameters
+    get_params( parser ) # getting all of the possible parameters
+    params = parser.parse_args() # parameters!
+
+    # reading the intitial profile
+    if params.model_type.lower() == 'mesa':
+        if params.model_header > 0:
```
AstroBarker: Looks like the header bound was updated for KEPLER, but not MESA, unless I missed something?

scripts/python/stellar/stellar.py
```diff
@@ -0,0 +1,211 @@
...
+            prof_new = np.zeros((rad_new.size, 10))
+            prof_new[:, 0] = rad_new
+
+            for i in range(1, prof_raw.shape[1]):
+                prof_new[:, i] = subsample_depr( prof_raw[:, i], rad_depr, factor)
+
+            prof_raw = prof_new
+
+        # weighted moving average
+        if params.wma_all:
```
AstroBarker, line 159: This should probably be a for loop.

scripts/python/stellar/convert.py
```diff
@@ -0,0 +1,163 @@
...
+    prof_conv = np.copy(
+        prof
+    )  # making a fresh copy of the profile so we don't overwrite the physical units.
+
+    rhoc = prof_conv[0, 1]  # central density
+    M0 = 1.0 / (rhoc**0.5) * ((c**2.0 / G) ** 1.5)  # characteristic mass
+    R0 = (G * M0) / (c**2.0)  # characteristic radius
+
+    # --- primitive quantities
+    prof_conv[:, 0] *= (c**2.0) / G / M0  # radius
```
AstroBarker, line 59: Not strictly needed now, but it might be more clear and future proof for "magic" indices referring to variables to be an enum or other named variable.

logalexw: oh good idea, I can update that on a later pr (probably when working on adding homologous/tov)

scripts/python/stellar/stellar.py
logalexw: nope, a misread on my end!

AstroBarker (approved):
Looking good now, thanks @logalexw !

Issue #246, opened 2026-07-17
getting ccsne progenitors to collapse


logalexw:
The journey to CCSNe in phoebus is ongoing, so @carlnotsagan and I are crowdsourcing ideas and/or suggestions for why the progenitor problem is (still) not working correctly. 

There's been quite a lot of troubleshooting and testing done to try and narrow down the issue (physics? computation? a typo in the codebase?) with mixed success (see below for the full list), but we're still seeing progenitors that initialize and run in phoebus without actually evolving any of the hydrodynamic variables. Initialization is at least correct, but after the zeroth timestep physics does not progress as expected at all. Currently models seem to oscillate in velocity without actually beginning infall (even at late times) while all other hydro quantities stall or evolve very, very slowly.

Ideas and questions are very welcome (as I'm running out of things that I think are possibly wrong)!

**Fig. 1.** Evolution of a KEPLER progenitor in phoebus (with an SFHo EOS, monopole GR, and lightbulb radiation).
<img width="2979" height="1577" alt="Image" src="https://github.com/user-attachments/assets/51666fbd-73a7-4c79-a3cd-87501ea24511" />

---
## What have we tried?

Major tests we've tried so far include (but are definitely not limited to):
- ensuring a uniform velocity gradient from $r = 0$ outward (e.g. all velocities should be negative in innermost zones; a suggested check by Jared Goldberg to avoid any weird edge effects or perturbations)
- broadening (and re-shrinking?) the radial domain of the initial profile (we started around ~1e9 cm and expanded to include a more `GR1D`-like cut at $\rho = 2 \cdot 10^3$, which puts our simulation domain out to around 5e9 cm.
- taking a FLASH-style approach to input (instead of extrapolating the input file to $r = 0$ during the pre-phoebus pipeline, we do an inner zone velocity fixup and let phoebus handle the interpolation at initialization)
- initializing and running the progenitor with an ideal gas EOS (same setup as `homologous.pin`)
- adjusting outer boundary conditions to a more realistic setup (e.g. changing `ox1_bc = reflect` to `ox1_bc = gr_outflow`
- checking the ceilings/floors. and then checking the floors again.
- enabling both lightbulb radiation and the Liebendörfer deleptonization scheme in `<rad>` so that Ye would evolve correctly
[...]

PR #248, opened 2026-08-10
adding full liebendorfer deleptonization

logalexw:
## PR Summary

After a bit of examination with the homologous collapse tests detailed in #246, I noticed that the deleptonization prescription implemented in the current code base only used the [Liebendörfer 2005](https://ui.adsabs.harvard.edu/abs/2005ApJ...633.1042L/abstract) method for updating Ye (i.e. how we do the final calculation of dYe), but with a sixth-order polynomial fit to a pre-calculated fit of Ye evolution based on log density. See fig. 1 below for an example of how this changes initial Ye evolution.

I've added some tweaks to how we handle deleptonization in the input decks (e.g. `do_delep` instead of `do_liebendorfer`, with appropriate method selection and error handling), as well as the full deleptonization prescriptions (method and fits) provided in Liebendörfer (G15 and N13). I'll include more figures for the homologous and progenitor tests run with the new delep methods shortly.

As a tiny bonus, there's also a typo fix included for a single variable name I missed in #245.

---

**Fig. 1.** Example implementation of the current log-density deleptonization fit included in phoebus, in comparison to the two prescriptions from Liebendörfer. Note that the changes in Ye here have greater amplitiude than the final phoebus implementation-- we update with dYe/dt in-code, where this figure just shows dYe.

Unknown-7

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/inputs/progenitor.pin
+++ b/inputs/progenitor.pin
@@ -138,10 +138,11 @@ gam0_ceiling = 1.0e4
 # simple example for lightbulb case
 <radiation> 
 cfl = 0.5 
-method = cooling_function
+method 			= cooling_function
 do_lightbulb 	= true
-do_liebendorfer	= true
-do_gain_calc = false # defaults to false if not set
+do_delep		= true
+delep_method 	= liebendorfer_g15 # options are rho_fit, liebendorfer_g15 (default), and liebendorfer_n13
+do_gain_calc 	= false # defaults to false if not set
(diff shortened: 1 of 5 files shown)


Yurlungur (approved):

AstroBarker (approved):
This looks good to me, thanks for this! We are surely getting closer.

PR #249, opened 2026-09-02
also adding liebendorfer entropy 

logalexw:
## PR Summary

After conversation with @carlnotsagan and @AstroBarker, it was suggested that we consider adding the entropy evolution that's implemented in [Liebendörfer 2005](https://ui.adsabs.harvard.edu/abs/2005ApJ...633.1042L/abstract) (**ref.** eq. 5 for the full prescription) in addition to the $Y_e$ deleptonization that was originally added in #248.

The version that's now in phoebus is given by: $dS = -d Y_e ( \mu_e - \mu_n + \mu_p) / T$.

Since we don't actually track entropy in phoebus, I thought it would be best to do this update through primitives (e.g. updating our temperature at constant $\rho$) with the help of singularity-EOS. I took chemical potentials directly from the EOS with singularity, and then implemented a quick root find to estimate the temperature that should produce the calculated change in entropy at a constant value of $\rho$.

This does seem a bit fast and lose thermodynamically, so if people have suggestions for a better implementation they are very welcome! Figures from the tests are coming very soon.

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/radiation/cooling_function.cpp
+++ b/src/radiation/cooling_function.cpp
@@ -306,8 +306,10 @@ TaskStatus CoolingFunctionCalculateFourForce(MeshData<Real> *rc, const double dt
               Real garbage;
 
               // all our const values
-              const Real E_nu = rad->Param<Real>("delep_Enu") *
+              const Real E_nu = rad->Param<Real>("delep_entropy_Enu") *
                                 MeVToErg; // escape energy param., in MeV
+              const Real rho_trap =
+                  rad->Param<Real>("delep_entropy_rho"); // trapping density, g/cm^3
               const Real T0 =
                   v(b, p::temperature(), k, j, i) * temperature_conversion_factor;
               const Real rho0 = rho;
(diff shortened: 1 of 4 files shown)


src/radiation/cooling_function.cpp
```diff
@@ -284,9 +292,62 @@ TaskStatus CoolingFunctionCalculateFourForce(MeshData<Real> *rc, const double dt
...
+                  return eos_sc.EntropyFromDensityTemperature(rho0, T, lambda) -
+                         (S0 + dS);
+                };
+
+                root_find::RootFind root_find;
+                dT = root_find.regula_falsi(target, eos_sc.TMin(), eos_sc.TMax(),
+                                            epsilon * T0, T0);
+
+                // final temperature update - should this be done after lightbulb?
+                v(b, p::temperature(), k, j, i) = dT / temperature_conversion_factor;
```
Yurlungur, line 348: I know this feels a bit hacky, but I do think it's thermodynamically sound. This is what I would suggest doing also.

Yurlungur (approved):
Too bad (though maybe a relief) that this doesn't make a big impact. Happy to include it though.

PR #250, opened 2026-09-12
adding utils for the progenitor problem

logalexw:
## PR Summary

Now that we're successfully getting progenitors to collapse, reach bounce, and explode (see closing comment in #246), some useful ccsne-related utilities are needed. In addition to some small updates to the progenitor pre-processing pipleine (`stellar.py`; #245) that are useful for setting an appropriate simulation time, I'm also adding parameters to track bounce-related diagnostics (if it occurs, density, entropy, etc.) and generate a new summary file.

As always, recommendations and suggestions are welcome!

## PR Checklist

- [x] Explain what you did.

### DIFF
--- a/src/progenitor/progenitordata.cpp
+++ b/src/progenitor/progenitordata.cpp
@@ -122,6 +122,15 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   params.Add("S_adm_dev", S_adm_dev);
   params.Add("Srr_adm_dev", Srr_adm_dev);
 
+  // criterion for bounce in ccsne (O'Connor & Ott 2010)
+  params.Add("bounce_density", Constants::BOUNCE_DENS);
+  params.Add("bounce_entropy", Constants::BOUNCE_ENTR);
+  // set these if bounce occurs
+  params.Add("bounce_density_actual"; -1.0); 
+  params.Add("bounce_entropy_actual"; -1.0); 
+  params.Add("bounce_time"; -1.0);
+  params.Add("post_bounce", False);
+
   params.Add("outside_pns_threshold", outside_pns_threshold);
   params.Add("inside_pns_threshold", inside_pns_threshold);
   params.Add("net_heat_threshold", net_heat_threshold);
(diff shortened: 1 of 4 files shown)


logalexw:
This one took a bit of digging to figure out, but there's now a small post-cycle diagnostic that runs for the progenitor problem only (`src/progenitor/progenitordata.cpp:GetProgenitorState`) and checks a) if the progenitor has reached bounce and b) handles a small bit of i/o into a bounce diagnostics file if it has.

I'm fairly sure all of the reducers should work correctly in 3d, but if there are any obvious missteps I'd appreciate a second set of eyes!

> This uses the same common bounce criterion as GR1D, s.t. $\rho \geq$ 2e14 g/cc indicates bounce has occured.

I also added a new condition to `<progenitor>` so that users can indicate if a model is post-bounce or not in the input deck. There's a few small tweaks to the *.info file output by the stellar pre-processing pipeline to make unit conversions more convenient as well.

src/progenitor/progenitordata.cpp
```diff
@@ -169,4 +188,86 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+      progen->UpdateParam<bool>("post_bounce", true);
+
+      // output data file, in code units and cgs units
+      FILE *fout;
+      fout = fopen("bounce.dat", "w"); // questionable naming...
+      fprintf(fout, "%-20s\n", ">> bounce reached!");
+      fprintf(fout, "%-16s  %.14e\n", "time", bounce_time);
+      fprintf(fout, "%-16s  %.14e\n", "central density", max_density);
+      fprintf(fout, "%-16s  %.14e\n", "central entropy", min_entropy);
+      fclose(fout);
```
Yurlungur, line 256: this is not MPI safe.  So you need an MPI reduce before you output so you get the real max density for example.

You could also just print to terminal and stash these quantities in params so they get output to file automatically, instead of writing to an ascii file. I personally feel that's a little nicer.

Yurlungur (commented):
Nice little improvements.

